PR #13, opened 2020-01-07
Add mqtt5 to index

nschejtman:


README.md
```diff
@@ -10,6 +10,7 @@ This repository contains the specifications for each AsyncAPI protocol binding.
 * [JMS binding](./jms)
 * [Kafka binding](./kafka)
 * [MQTT binding](./mqtt)
+* [MQTT5 binding](./mqtt5)
```
fmvilas: Is the protocol that different from previous versions of MQTT? Can't we add the new stuff to the existing binding and make it clear that's only for v5?

nschejtman: Frankly I’ve never used MQTT. I’m working on a parser for AsyncAPI 2.0 and I’ve seen MQTT5 is listed as one of the possible bindings but is missing in this index.

fmvilas: Oh yeah, you're right. Sorry for the dumb question. Thanks for fixing it, I'll merge it.

PR #16, opened 2020-04-01
FIFO & Standard Queue Config

Paul-T-AU:
Adding config to control ordering & deduplication when using FIFO queues.


Paul-T-AU:
@fmvilas As discussed the cleaned up ordering & dedupe binding for SQS.

Paul-T-AU:
Proposing as an extension so closing this request.

PR #42, opened 2021-01-02
fix: description for MQTT QoS for Operation Binding Object

Lazzaretti:
**Description**

- The field `qos` in the MQTT Bindings is described quite differently than in the specification:
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718099

**Related issue(s)**

### DIFF
--- a/mqtt/README.md
+++ b/mqtt/README.md
@@ -68,7 +68,7 @@ This object contains information about the operation representation in MQTT.
 
 Field Name | Type | Applies To | Description
 ---|:---:|:---:|---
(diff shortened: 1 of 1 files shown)


derberg:
@Lazzaretti lgtm, your description is much better 😅 but with your change you also removed the important part `| Publish, Subscribe |` and now the table doesn't indicate to what operations can this binding be applied

Lazzaretti:
@derberg thanks for the feedback, I fixed the broken table by re-adding the missing cell

derberg (approved):

Issue #41, opened 2020-12-18
Specify how Avro schema ID is serialized on Kafka


lbroudoux:
#### Reason/Context

AsyncAPI allows to reference Avro schema used for serializing / deserializing messages on a topic and parser has a sample on [how to express that with a schema being already into a registry](https://github.com/asyncapi/avro-schema-parser#usage-with-confluent-schema-registry). Nice !

When it comes to serializing Avro data to a Kafka topic, you usually have 2 options :
* The « old-fashioned one » that is about putting raw Avro binary representation of the message payload,
* The « modern one » that is about putting the Schema ID + the Avro binary representation of the message payload (see [Schema Registry: A quick introduction](https://www.confluent.io/blog/kafka-connect-tutorial-transfer-avro-schemas-across-schema-registry-clusters/).

As you have guessed, using one or the other of these options has impact on the consumers ! But as of today, consumer has no mean to know which one of the serialization mode is used.

Moreover, it appears that different SerDes libraries (Confluent, Apicurio and IBM as examples) have different way of serializing the Schema ID. Confluent is using first byte of payload to encode the schema ID whereas Apicurio and IBM SerDes are also able to put the ID into headers.

After exposing this issue on the [AsyncAPI Slack](https://asyncapi.slack.com/archives/C34F2JV0U/p1608289028063000), there's also concerns on how to retrieve Schema Registry endpoint URL. As we already have Server information with specific bindings info into AsyncAPI, it can be considered as legit - in the case a Schema Registry is needed to deserialize message - to also allow specification of the Schema Registry URL associated to Server. 

#### Description

I propose to add a new Kafka specific binding attribute like at the `Channel` level to specify how the Schema ID is encoded. We could call it `schemaIDLocation` with possible values being `header` or `payload`. If binding attribute is missing then it simply means that Schema info is not provided. Depending on location value you may switch to different SerDes implementation for consumer generation or configuration

I also propose to add a new Kafka specific binding attribute at the `Server` level to specify the Schema Registry URL. Simply name it `schemaRegistryUrl`.

dalelane:
I had a go at illustrating what this would look like. It got a bit long, so instead of putting it in a super-long comment here, I've put it in a stand-alone gist at https://gist.github.com/dalelane/3931c17b14c51fa4a1cf25496237d188

nictownsend:
I believe - binary/json can be inferred from the message binding content type? application/json or application/octet-stream?

dalelane:
[...]

Issue #40, opened 2020-12-18
Specify Avro encoding approach


dalelane:
There is a need to describe the Avro encoding when using Apache Avro as the payload schema format.

For context, when using an Avro schema to serialize data, a Kafka producer can choose between [two ways of encoding the data](https://avro.apache.org/docs/current/spec.html#Encodings):

- json - the data is still human-readable which is sometimes useful
- binary - the encoded data is smaller and can be faster to deserialize

A Kafka consumer needs to know which encoding mechanism was used to be able to deserialize the messages. 

One of the most commonly used Kafka serializer libraries, Confluent's `io.confluent.kafka.serializers.KafkaAvroSerializer`, only [supports binary encoding](https://github.com/confluentinc/schema-registry/blob/master/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroSerializer.java#L103), so I think it's reasonable to assume binary as a default where not specified, and not try to treat this as a required field. 

But it should be possible to specify when json encoding is being used with other serdes libraries [such as Apicurio's](https://github.com/Apicurio/apicurio-registry/blob/0195ace9f4929361114e215787bd0c7d916fdef2/utils/serde/src/main/java/io/apicurio/registry/utils/serde/AvroEncoding.java#L20-L40)

([related Slack question](https://asyncapi.slack.com/archives/C34F2JV0U/p1608292454066500))

Fran's suggestion in Slack was that capturing this in the operation binding would allow this to be specified at the topic level.

dalelane:
I had a go at illustrating what this would look like. It got a bit long, so instead of putting it in a super-long comment here, I've put it in a stand-alone gist at https://gist.github.com/dalelane/3931c17b14c51fa4a1cf25496237d188 

(I also addressed the related issue #41 as part of the same exploration)

derberg:
@dalelane is it Kafka specific only? why not on schemaFromat level?

dalelane:
Yeah, that's a fair point. I've only heard of this being an issue in Kafka, but in theory there's no technical reason why it only could apply to Kafka

derberg:
at the end you just need to specify in the document what content type is sent over the wire right? like `avro/binary` and the schema format would still be json or yml, or?

dalelane:
@derberg When you're using a schema registry (which is the most common approach when Kafka developers are using Avro schemas), developers will specify the ID and version for the schema they're using. There are lots of different places that this can be specified. 

[...]

PR #55, opened 2021-02-23
docs: update Kafka bindings to describe schema usage

dalelane:
Adding enough information about how schemas are being
used in Kafka messages to enable an application that
consumes messages serialized using schemas to be
implemented purely from an AsyncAPI spec.

Also introduces two new common, reusable definitions
that can be used as message traits, to make it easier
to quickly specify headers used to capture schema
details.

Contributes to: asyncapi/bindings#40
Contributes to: asyncapi/bindings#41

Signed-off-by: Dale Lane <dale.lane@uk.ibm.com>


dalelane:
I've put an illustration of how this could be used in https://gist.github.com/dalelane/3931c17b14c51fa4a1cf25496237d188

kafka/traits/apicurio-v0.0.1.yaml
```diff
@@ -0,0 +1,14 @@
+name: apicurio
```
fmvilas, line 1: Can we include these examples inside the README.md file? Let's try to make the spec a single document as much as we can, so it's easier to work with it.

dalelane: The motivation (and I can't take the credit for it as it was @lbroudoux's idea, although I thought it was really good) for putting them in separate files was to enable people to refer to common sets of serdes headers as message traits.

e.g. https://gist.github.com/dalelane/3931c17b14c51fa4a1cf25496237d188#how-this-could-be-described-in-asyncapi-5 

```yaml
channels:
  my.topic.name:
    ...
    subscribe:
      ...
      message:
        schemaFormat: "application/vnd.apache.avro;version=1.9.0"
        contentType: "application/octet-stream"
        traits:
          - $ref: https://github.com/asyncapi/bindings/raw/master/kafka/traits/apicurio-v0.0.1.yaml
```

It makes a nice convenient short-hand to avoid having to copy/paste these standard serdes headers into many specs.

fmvilas: 🤔 Ok. Then let's put them inside an examples directory maybe? Sounds like they are part of the specification.

lbroudoux: I think this mechanism could also be used to integrate smoothly with CloudEvents.io. This will encourage building on and not limiting to this standard. Sometimes people tend to think that "if it's cloudevents.io compatible, it's sufficiently contractualized ..."

More generally, let's imagine having some "off-the-shelf" traits holding the headers for standards like CloudEvents and its different protocol bindings - but think also about OpenBanking async operations, Fiware, ....

These could be located into a dedicated repo where community can make PR contributing their own traits.

kafka/README.md
```diff
@@ -6,16 +6,36 @@ This document defines how to describe Kafka-specific information on AsyncAPI.
...
-This object MUST NOT contain any properties. Its name is reserved for future use.
+This object contains information about the server representation in Kafka.
 
+##### Fixed Fields
 
+Field Name | Type | Description | Applicability [default] | Constraints
+---|:---:|:---:|:---:|---
+`schemaRegistryUrl` | string (url) | API URL for the Schema Registry used when producing Kafka messages (if a Schema Registry was used) | OPTIONAL | -
+`schemaRegistryVendor` | string | Identifies the Kafka serdes library that should be used (e.g. `confluent`, `apicurio`, or `ibm`) | OPTIONAL | MUST NOT be specified if `schemaRegistryUrl` is not specified
+`schemaRegistryAvailable` | boolean | Specifies if the Schema Registry identified in `schemaRegistryUrl` is available for use by consumers of the AsyncAPI spec | OPTIONAL [true] | MUST NOT be specified if `schemaRegistryUrl` is not specified
```
lbroudoux, line 24: I have some doubt with `schemaRegistryAvailable` as explained in https://github.com/asyncapi/bindings/issues/41#issuecomment-785898884.

I think we still need a more generic `useSchemaRegistry` flag that is tied to the operation level. For me it brings information to the producer that it should integrate with a registry - and this very early in the lifecycle of the API, long before knowing where it will land in terms of servers.

lbroudoux (commented):
Thanks @dalelane for great work. In my opinion, there's still the need for `useRegistry` flag dissociated from servers.

kafka/README.md
dalelane: I think there are two issues.

1) Describing the data that is in the message
2) Describing other servers where schema information is stored


**(1) Describing the data that is in the message**

I was trying to address this in the message binding object, using the `schemaIdLocation` value.

If the message was produced with a Confluent serdes, then the payload will have the schema id and version inserted at the start. (schemaIdLocation = payload)

(This means you could consume using a Confluent serdes library, which knows how to handle that. Or you can consume yourself by swallowing the beginning bytes before attempting the Avro deserialization)

If the message was produced in, well.... pretty much any other way, then the payload will just be serialized Avro data. (schemaIdLocation = header or none)

(This means you can consume it with a non-Confluent serdes library if you have access to a schema registry, or you can just Avro deserialize the message contents yourself)  


**(2) Describing servers where schema information is stored**

As this isn't an attribute of the data on the topic, I felt like that the server bindings was the best place to put it. 

As it is information about the schema registry that stores schema data (and whether or not people are allowed to connect to it) and not an attribute of the data on the topic, I didn't think it belonged in the message binding. 

Similarly, with `schemaRegistryAvailable` (or `useSchemaRegistry` or whatever we end up naming it) as it is information about the schema registry, and whether or not people are allowed to connect to it, it's still not an attribute of the data on the topic, I didn't think it belonged in the message binding. 

That leaves putting it in the operation, channel, or server bindings as the most obvious places.

The benefit of putting it in at the server level is that it means you don't have to duplicate it. You declare it once - "here's my schema registry and you can use it" or "here's my schema registry, but sorry it's private and you can't use it". 

The drawback of putting it at the server level is that if you have some topics where you would allow people to use the schema registry, and other topics where you won't, then you'd need to declare them in separate AsyncAPI specs. (_In my opinion, that seems like an unlikely edge case, so that seems okay to me as a workaround_)

The main benefit I can think of for putting it in the operation binding would be if you treated producing and consuming differently - so you could say people are allowed to use a schema registry for sending messages but not receiving them (or vice versa). Are there other benefits I'm missing?

I can't think of a strong reason for putting it in the channel binding. As I say, it seems unusual to say a schema registry is available for use for one schema but not for another. And if that's really the case, then you could address that in separate specs. 

Dunno, to be honest - I don't have a use case in mind for this field, so I'm very much guessing here.

lbroudoux: I'm totally aligned with use-case (1) 😉 

**Some comments regarding use-case (2)**

Agree with you on the location : either server or operation (not message related and can't see any reason it would be channel)

A concern I have with putting this into `servers` is that I met several customers - that are currently using OpenAPI v3 and are looking at AsyncAPI - that are not using `servers` or are using them very late in the API lifecycle... When architects or domain experts at these customers are designing an API, they indeed do not know (and even do not care) on what servers it will be hosted! And I can easily understand that!

Sometimes, the `server` part is later completed as the API implementation is going through QA and Production - usually when it's finally published onto the Developer Portail of the in-house API Management solution - but it's not even a mandatory step.

Now translating to AsyncAPI and our Kafka concerns, I think the having this information of "will we use a registry ?" is very important at design time - because we may have some extra work designing Avro schema as well - and very valuable for developers to come. Moreover I think it's totally orthogonal with the fact of knowing the registry server (`schemaRegistryUrl`) and the fact of knowing if we are going to make it available (`schemaRegistryAvailable`).

So to summarize things up:
* Totally agree with `schemaRegsitryUrl` and `schemaRegistryAvailable` at the `server` level. The `server` block may be not specified at first and then added later in the lifecycle,
* Still thinking we need `useSchemaRegistry` independently at the `operation` level to tell developers, producers and consumers that there's possibly an extra schema to design / publish / retrieve

NB: Another approach could be to deduce this usage of schema registry from the presence of `schemaIdLocation`. If it has a value that is not `none` then it implies that we're using a schema registry ... I think it should do the trick but for me it's far less explicit.

dalelane: That's useful insight, thanks - as I said, I was guessing because I didn't have a use case in mind, so it's useful to have a more concrete use case as guidance.

I've added `useSchemaRegistry` to the operation binding.

I'm still a little unsure about having both `schemaRegistryAvailable` and `useSchemaRegistry` though - mainly because  a contradictory spec that sets `schemaRegistryAvailable` to `false` and `useSchemaRegistry` to `true` would be telling developers they should use a schema registry, but they're not allowed to connect to it.

What do you think?

kafka/traits/apicurio-v0.0.1.yaml
dalelane: @fmvilas @lbroudoux If we think that this better belongs in a different repo, do you think I should remove the examples from this PR, as I'd be reluctant to move them after people start referring to the URLs for trait definitions here?

dalelane: Removed these from this PR as I think we can consider them separately to the proposed changes to the bindings

kafka/README.md
dalelane: I've removed it for now, as we can always add it in a future PR.

kafka/traits/apicurio-v0.0.1.yaml
fmvilas: Sorry, I missed your previous comment. Yeah, I agree examples can be placed somewhere else. For now, we're putting them together at https://github.com/asyncapi/asyncapi/tree/master/examples/2.0.0.

kafka/README.md
```diff
@@ -6,16 +6,34 @@ This document defines how to describe Kafka-specific information on AsyncAPI.
...
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
+This object contains information about the server representation in Kafka.
 
+##### Fixed Fields
 
+Field Name | Type | Description | Applicability [default] | Constraints
+---|:---:|:---:|:---:|---
+`schemaRegistryUrl` | string (url) | API URL for the Schema Registry used when producing Kafka messages (if a Schema Registry was used) | OPTIONAL | -
+`schemaRegistryVendor` | string | Identifies the Kafka serdes library that should be used (e.g. `confluent`, `apicurio`, or `ibm`) | OPTIONAL | MUST NOT be specified if `schemaRegistryUrl` is not specified
```
lornajane, line 23: I'm not sure what the criteria are for being considered a registry vendor are, but could we include `karapace` here please? (https://karapace.io/, open source schema registry tool for Kafka)

dalelane: Yeha, absolutely (I only included those examples because those are ones I've used, but it's absolutely not meant to be exhaustive)

derberg:
where are we with this PR? looks good to me

M3lkior:
it is a good start ! some users (like me) needs to provide extra informations about topics naming strategy like this (confluent specific)

 ```
x-key.subject.name.strategy:
            type: string
            description: |
              We use the **RecordNameStrategy** to infer the messages schema.
              Use `io.confluent.kafka.serializers.subject.RecordNameStrategy` for your consumer.
          x-value.subject.name.strategy:
            type: string
            description: |
              We use the **RecordNameStrategy** to infer the messages schema.
              Use `io.confluent.kafka.serializers.subject.RecordNameStrategy` for your consumer.
```

So maybe it will be nice to allow bindings extension by `x-` prefixes

A Slack community discussion is started here : https://asyncapi.slack.com/archives/C34F2JV0U/p1618495734389300

**Comment by derberg** [comment]
@M3lkior where does it say extensions in bindings are not allowed? I mean that additional properties are not allowed? 

when you do
```
    bindings:
      kafka:
        x-test: test
```

all is good, but for now of course because of the fact that bindings do not have schemas

**Comment by M3lkior** [comment]
> @M3lkior where does it say extensions in bindings are not allowed? I mean that additional properties are not allowed?
> 
> when you do
> 
> ```
>     bindings:
>       kafka:
>         x-test: test
> ```
> 
> all is good, but for now of course because of the fact that bindings do not have schemas

i try to set bindings with `x-` prefixes, but they are not displayed or returned by the `bindings()` method

**Comment by derberg** [comment]
@M3lkior you sure? I recently updated playground with latest HTML where you contributed bindings info rendering and I can see the extension is rendered 👇🏼 

Screenshot 2021-04-20 at 09 40 19

If spec doesn't say extensions are not allowed, then we are good. Bindings object can be extended and protocol-specific binding objects too. There is some work on schemas started https://github.com/asyncapi/bindings/pull/60

Can I suggest you create a separate issue in the spec repo to clarify if we should not have explicitly described in the spec that specific bindings must always be extendable? Instead of having it as a decision per binding?

**Comment by M3lkior** [comment]
> @M3lkior you sure? I recently updated playground with latest HTML where you contributed bindings info rendering and I can see the extension is rendered 👇🏼
> 
> Screenshot 2021-04-20 at 09 40 19
> 
> If spec doesn't say extensions are not allowed, then we are good. Bindings object can be extended and protocol-specific binding objects too. There is some work on schemas started #60
> 
> Can I suggest you create a separate issue in the spec repo to clarify if we should not have explicitly described in the spec that specific bindings must always be extendable? Instead of having it as a decision per binding?

yes sorry : 

image

(by the way, there is a rendering issue with too long text in the declaration)

**Comment by derberg** [comment]
@M3lkior no worries. Fix for the bug https://github.com/asyncapi/html-template/pull/183

**Comment by dalelane** [comment]
> where are we with this PR? looks good to me

@derberg Sorry, been super busy recently and this one got away from me. 

I've changed the state to Draft to reflect that there are a few bits of feedback that I need to consider and reflect in the PR:
- Apicurio pointed out that they have different approaches to encoding schema info in the body that I'm not adequately capturing
- @M3lkior's [pointed out](https://github.com/asyncapi/bindings/pull/55#issuecomment-822608500) that I haven't captured Confluent strategies

**Comment by lbroudoux** [comment]
Hi there!

Had a review this morning and I'm struggling with the `useSchemaRegistry` and `schemaIdLocation` properties when this later value is `none`.

From the binding README, definition of `useSchemaRegistry` is clear but the examples from @dalelane Gist confuse me as they seems to favor the `schemaIdLocation=none` approach. Especially the first one here: https://gist.github.com/dalelane/3931c17b14c51fa4a1cf25496237d188#binary-encoded-avro

Also the default value for `schemaIdLocation` is `none` which - as I understand from the Gist - means `useSchemaRegistry` is equivalent to `false` - but this one has no default value specified.

I think we should solve this ambiguity and I see 2 options at the moment:

1. Keep both properties and have `useSchemaRegistry` default to `false` and restrict `schemaIdLocation` to either `payload` - the default because the most common -  or `header`,

2. Remove the `useSchemaRegistry` and just keep `schemaIdLocation` with `none` as default, + the `payload` and `header` values.

From my perspective, I prefer option 1. as I find it more easy to use for non-expert of SerDes library and more straightforward for most common cases. That way, you may have something like:

```yaml
channels:
  user-signedup:
    publish:
      bindings:
        kafka:
          groupId:
            type: string
            enum: ['myGroupId']
          clientId:
            type: string
            enum: ['myClientId']
          bindingVersion: '0.2.0'
```

and this is telling that this operation/channel is not using any registry. Then we can enable registry usage explicitly:

```yaml
channels:
  user-signedup:
    publish:
      bindings:
        kafka:
          groupId:
            type: string
            enum: ['myGroupId']
          clientId:
            type: string
            enum: ['myClientId']
          useSchemaRegistry: true
          bindingVersion: '0.2.0'
```

and this mean that a Schema registry is used and consumer will have to pick the SerDes implementation corresponding to registry vendor. The Schema ID Location will thus be the `payload` of the message. Then we can also refine things adding:

```yaml
channels:
  user-signedup:
    publish:
      bindings:
        kafka:
          groupId:
            type: string
            enum: ['myGroupId']
          clientId:
            type: string
            enum: ['myClientId']
          useSchemaRegistry: true
          bindingVersion: '0.2.0'
      message:
        bindings:
          kafka:
            schemaIdLocation: 'header'
            bindingVersion: '0.2.0'
```

and this is necessary only is we're not sticking with most common usage.

What do you think?

nictownsend:
@lbroudoux I think there are two use cases this is looking to handle:

1. Inform API developers to enable them to pick the right vendor serdes
2. Inform API developers how to bypass if schema registry is unavailable

For use case 1 - `schemaRegistryURL + schemaRegistryVendor` is sufficient to pick a SERDES library. `useSchemaRegistry` flags if the SERDES library should be used for the operation (if not, assumption is message has been produced without using a SERDES library (e.g basic producer using AVRO). 

Missing: 
- Naming strategies in each operation allow additional SERDES config for [Confluent](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#subject-name-strategy) + [Apicurio](https://www.apicur.io/registry/docs/apicurio-registry/2.0.0.Final/getting-started/assembly-using-kafka-client-serdes.html#registry-serdes-concepts-strategy-registry). These maybe just need to be class names so that the key can be vendor agnostic (`naming/lookupStrategy`)?
- [Apicurio ID encoding types](https://www.apicur.io/registry/docs/apicurio-registry/2.0.0.Final/getting-started/assembly-using-kafka-client-serdes.html#registry-serdes-types-avro-registry)

For use case 2 - to deserialize from the payload without using a SERDES library you need to know:
- `schemaIdLocation` - so that you know if you have to skip any bytes (Apicurio looks for a magic byte to automatically determine if in header or not - confluent only use payload, Event Streams uses header only). Maybe this just needs to be `schemaIdInPayload = true/false`
- `schemaRegistryVendor` - so that for each wire format e.g [Confluent](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format), [Apicurio](https://www.apicur.io/registry/docs/apicurio-registry/2.0.0.Final/getting-started/assembly-using-kafka-client-serdes.html#registry-serdes-concepts-serde-registry) _(find in page 'magic byte')_ - you know how many bytes to skip (magic + schema id)
- Apicurio ID encoding type to skip the right number of bytes for the schema ID

I think this means that there are two sets of properties - with overlap but no conflict. You could combine the two sets to describe both use cases on the same spec without any clashes (though for code generation purposes I guess you'd generate using SERDES if all required SERDES properties are described).

I think the only change to the constraint is that `schemaRegistryVendor` is required if either `schemaIdInPayload` or `schemaRegistryURL` are supplied.

Does that clear up the ambiguity by separating properties into the two different concerns?

lbroudoux:
Thanks @nictownsend : this makes things clearer.

However, I'd thought that `schemaIdLocation` is still useful for use case 1 - enabling developers to correctly setup SERDES properties. I especially think to the Apicurio usage of headers for ID.

```java
props.putIfAbsent(SerdeConfig.ENABLE_HEADERS, "true")
```
Is it correct?

+1 for the missing lookup strategies!

nictownsend:
> Thanks @nictownsend : this makes things clearer.
> 
> However, I'd thought that `schemaIdLocation` is still useful for use case 1 - enabling developers to correctly setup SERDES properties. I especially think to the Apicurio usage of headers for ID.

Good point, yes for Apicurio case - `schemaIdLocation` is needed for producer configuration (I had stupidly focused on consume only).

So do we then have the following properties:
- `schemaRegistryURL` - url for the registry (use case 1 only)
- `schemaRegistryVendor` - SERDES library to use (use case 1 only)
- `schemaIdLocation` - where to find the ID (and thus using this property means `useSchemaRegistry` is no longer needed as use of this is stating "this operation uses SERDES") (both use cases)
- `schemaIdPayloadEncoding` - length of bytes or vendor specific values (e.g confluent/apicurio-legacy/apicurio-new) (both use cases)
- `schemaLookupStrategy` - freeform `string` for any naming strategy class to use (use case 1 only) and clients default to the vendor default if not supplied

Only query is whether it can be assumed that every vendor uses a magic byte when writing schema to payload. If not, then `schemaRegistryVendor` becomes use case 2 as well.

PR #56, opened 2021-02-24
docs: add essential Kafka client security parameters to bindings

dalelane:
There are a huge number of Kafka client configs we could add to
server bindings, but I'm skeptical of the value of just copying
them all across. Most client config values have sensible
defaults that are suitable in most cases.
cf. https://kafka.apache.org/documentation/#consumerconfigs

Two values are essential for code generation with secured Kafka
clusters, however - identifying the security mechanism and
protocol to use. This commit adds these two most commonly-used
parameters to the Kafka server bindings. I suggest we start with
these and gauge usage before starting to add many more, as these
are enough to enable code generation for secured Kafka clusters.

Closes: asyncapi/asyncapi#466

Signed-off-by: Dale Lane <dale.lane@uk.ibm.com>

### DIFF
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -6,16 +6,36 @@ This document defines how to describe Kafka-specific information on AsyncAPI.
 
 ## Version
 
-Current version is `0.1.0`.
+Current version is `0.2.0`.
 
 
 <a name="server"></a>
 
 ## Server Binding Object
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
+This object contains information about the server representation in Kafka.
 
+##### Fixed Fields
 
+Field Name | Type | Description | Applicability [default] | Constraints
+---|:---:|:---:|:---:|---
+`saslMechamism` | string | Specifies the [`sasl.mechanism`](https://kafka.apache.org/documentation/#consumerconfigs_sasl.mechanism) config property that must be specified by Kafka clients connecting to this server. <br>Example values include `GSSAPI`, `PLAIN`, `OAUTHBEARER`, `SCRAM-SHA-256`, `SCRAM-SHA-512`  | OPTIONAL [`GSSAPI`] | -
+`securityProtocol` | string | Specifies the [`security.protocol`](https://kafka.apache.org/documentation/#consumerconfigs_security.protocol) config property that must be specified by Kafka clients connecting to this server. <br>Example values include `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT`, `SASL_SSL` | OPTIONAL [`PLAINTEXT`] | -
+<a name="serverBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. | OPTIONAL [`latest`]
+
+This object MUST contain only the properties defined above.
+
+##### Example
+
+```yaml
+servers:
+  production:
+    bindings:
+      kafka:
+        saslMechanism: 'SCRAM-SHA-512'
+        securityProtocol: 'SASL_PLAINTEXT'
+        bindingVersion: '0.2.0'
+```
 
 
 <a name="channel"></a>


smoya:
Hi @dalelane! Thanks for making this PR.
I'm wondering if we could add all those auth mechanisms as [Security Schemas](https://www.asyncapi.com/docs/specifications/2.0.0#securitySchemeObject) instead, so they can be reused. I see all those mechanisms are not **only** Kafka-oriented.

dalelane:
> I'm wondering if we could add all those auth mechanisms as [Security Schemas](https://www.asyncapi.com/docs/specifications/2.0.0#securitySchemeObject) instead, so they can be reused. I see all those mechanisms are not **only** Kafka-oriented.

Could you give me a bit more detail about how that'd work, please? Or maybe an example of what this would look like in security schemes? 
```
saslMechanism: 'SCRAM-SHA-512'
securityProtocol: 'SASL_PLAINTEXT'
```

smoya:
> Could you give me a bit more detail about how that'd work, please? Or maybe an example of what this would look like in security schemes?

We already have implemented one of the available SASL mechanisms: x509.
I think you could define a new schema for each mechanism, e.g. SASL SCRAM-SHA-512:

```yaml
type: saslScramSha512
```

Or create a global one for all SHA, which means we would have x509 duplicated, not so ideal I think.

```yaml
type: sasl
saslMechanism: ScramSha512
```

kafka/README.md
```diff
@@ -6,16 +6,36 @@ This document defines how to describe Kafka-specific information on AsyncAPI.
...
 <a name="server"></a>
 
 ## Server Binding Object
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
+This object contains information about the server representation in Kafka.
+
+##### Fixed Fields
 
+Field Name | Type | Description | Applicability [default] | Constraints
```
fmvilas, line 20: If the whole "Constraints" column is empty, shouldn't we just get rid of it entirely?

kafka/README.md
```diff
@@ -6,16 +6,36 @@ This document defines how to describe Kafka-specific information on AsyncAPI.
...
+---|:---:|:---:|:---:|---
+`saslMechamism` | string | Specifies the [`sasl.mechanism`](https://kafka.apache.org/documentation/#consumerconfigs_sasl.mechanism) config property that must be specified by Kafka clients connecting to this server. <br>Example values include `GSSAPI`, `PLAIN`, `OAUTHBEARER`, `SCRAM-SHA-256`, `SCRAM-SHA-512`  | OPTIONAL [`GSSAPI`] | -
```
fmvilas, line 22: Really the default value is `GSSAPI`? Isn't it `PLAIN`?

fmvilas: Also, shouldn't we have something like a `NONE` value to disable the saslMechanism?

kafka/README.md
dalelane: I was doing that to avoid the merge conflict if both this and https://github.com/asyncapi/bindings/pull/55 get approved

fmvilas:
I just saw Sergio's comment. Yeah, it's true it would make more sense to add this to Security Schemes. The way to proceed is similar. If something is missing on the Security Schemes Object, then feel free to add new keys, like `saslMechanism` and `securityProtocol`.

kafka/README.md
fmvilas: Oh, I see. Thanks for the explanation 👍

kafka/README.md
dalelane: I did that to mirror what Kafka does by default - which is GSSAPI

https://github.com/apache/kafka/blob/fc68c0fc9b1199b4eb63d9d486b012667c1d448f/clients/src/main/java/org/apache/kafka/common/config/SaslConfigs.java#L32

image

But I think adding a NONE option for when using `kafka` instead of `kafka-secure` might make sense, and make a better default.

fmvilas: Yeah. In any case, if we move this to the Security Schemes Object, we would not need the `NONE` value anymore. Sorry, for the messed-up review. I realized about the other comments after I made the review.

dalelane: no problem at all - it's been a useful thread, thanks to both you and @smoya !

nictownsend:
The problem with `x509` is that it becomes overloaded - or at least for Kafka - in that you can have SSL based authentication, and SASL based. But there's no `sasl x509` - so using `sasl` as a prefix on any new type would be misleading (unless `x509` is documented as shared between concepts).

smoya:
> The problem with `x509` is that it becomes overloaded - or at least for Kafka - in that you can have SSL based authentication, and SASL based. But there's no `sasl x509` - so using `sasl` as a prefix on any new type would be misleading (unless `x509` is documented as shared between concepts).

You are totally right @nictownsend. x509 would need to be a shared mechanism, in this case between TLS/SSL and SASL.
What if we just create one schema for each mechanism? 
- GSSAPI
- OAUTHBEARER (This is already available under the `oauth2` schema iirc)
- SCRAM-SHA-256 (Not sure if it's a good idea to reuse `userPassword` schema, same for the next one)
- SCRAM-SHA-512

dalelane:
That's been really useful, thanks @smoya @nictownsend 

I think where we end up is:

AsyncAPI: <br>server [`protocol`](https://www.asyncapi.com/docs/specifications/2.0.0#fixed-fields-4) | AsyncAPI: <br> security scheme [`type`](https://www.asyncapi.com/docs/specifications/2.0.0#fixed-fields-21) | meaning<br>encryption? | meaning<br>auth? | maps to Kafka client parameter<br>[`sasl.mechanism`](https://kafka.apache.org/documentation/#consumerconfigs_sasl.mechanism) | maps to Kafka client parameter<br>[`security.protocol`](https://kafka.apache.org/documentation/#consumerconfigs_security.protocol)
---|:---:|:---:|:---:|:---:|---
`kafka` |  | no | no | _unset_ | `PLAINTEXT` 
`kafka` | `plain` <sup>*</sup> | no | yes, using SASL/PLAIN | `PLAIN` | `SASL_PLAINTEXT`
`kafka` | `scramSha256` <sup>*</sup> | no | yes, using SASL/SCRAM | `SCRAM-SHA-256` | `SASL_PLAINTEXT` 
`kafka` | `scramSha512` <sup>*</sup> | no | yes, using SASL/SCRAM | `SCRAM-SHA-512` | `SASL_PLAINTEXT` 
`kafka` | `oauth2`  | no | yes, using OAuth | `OAUTHBEARER` | `SASL_PLAINTEXT` 
`kafka` | `gssapi` <sup>*</sup> | no | yes, using GSSAPI | `GSSAPI` | `SASL_PLAINTEXT` 
`kafka-secure` |  |  yes | no | _unset_ | `SSL`
`kafka-secure` | `plain` <sup>*</sup> | yes | yes, using SASL/PLAIN | `PLAIN` | `SASL_SSL`
`kafka-secure` | `scramSha256` <sup>*</sup> | yes | yes, using SASL/SCRAM | `SCRAM-SHA-256` | `SASL_SSL` 
`kafka-secure` | `scramSha512` <sup>*</sup> | yes | yes, using SASL/SCRAM | `SCRAM-SHA-512` | `SASL_SSL` 
`kafka-secure` | `oauth2`  | yes | yes, using OAuth | `OAUTHBEARER` | `SASL_SSL` 
`kafka-secure` | `gssapi` <sup>*</sup> | yes | yes, using GSSAPI | `GSSAPI` | `SASL_SSL` 
`kafka-secure` | `X509` | yes | yes, using mutual TLS | _unset_ | `SSL` 


<sup>*</sup> - a new security scheme type to be added

dalelane:
Closing as I think we've agreed this is better captured in security schemes, as described in https://github.com/asyncapi/bindings/pull/56#issuecomment-785834586 so I've opened https://github.com/asyncapi/asyncapi/pull/502 for this

PR #60, opened 2021-04-18
feat: json schemas for bindings

KhudaDad414:
**Description**

- this PR will add all of the JSON Schemas for different binding in different protocols.

**Related issue(s)**
See also [#507](https://github.com/asyncapi/spec/issues/507)


amqp/schema_channel.json
```diff
@@ -0,0 +1,61 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Channel Schema",
+  "type": "object",
+  "properties": {
+    "is": {
+      "type": "string",
+      "enum": ["queue", "routingKey"]
+    },
+    "exchange": {
+      "type": "object",
```
KhudaDad414, line 11: I couldn't figure out how to add Map type. should I change `object` to `any`?

derberg: `object` is exactly the way. You can always manually test schema against real object (taken from example) using https://www.jsonschemavalidator.net/

magicmatatjahu:
@KhudaDad414 Thanks for handling this task! I know that it's draft, but I advise you to keep the schemas in some folder, e.g. `amqp/json_schemas` and have a schema for messages, operations, etc. like` amqp/json_schemas/message.json`. What do you think about that? :)

KhudaDad414:
> @KhudaDad414 Thanks for handling this task! I know that it's draft, but I advise you to keep the schemas in some folder, e.g. `amqp/json_schemas` and have a schema for messages, operations, etc. like` amqp/json_schemas/message.json`. What do you think about that? :)

thanks for the input. I will put them inside a folder. any other advice regarding schema?

derberg:
I think we need this https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#L10-L15 on every schema file to make sure no additional props are added, but extensions (props with `x-`) are allowed.

@fmvilas on bindings you can find sentences that only props from given list are allowed but I guess we still want to have specific bindings extendable, right?

http/json_schemas/operation.json
```diff
@@ -0,0 +1,29 @@
...
+    },
+    "query": {
+      "type": "object",
+      "properties": {
+        "properties": {
+          "type": "object"
+        }
+      },
+      "required": ["properties"]
+    },
```
magicmatatjahu, line 22: In places when you have `Schema Object` type you should write reference to JSON Schema spec like:

```json
"query": {
  "$ref": "http://json-schema.org/draft-07/schema#"
}
```

Or use our AsyncAPI Schema -> https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#L359

http/json_schemas/message.json
```diff
@@ -0,0 +1,20 @@
...
+  "properties": {
+    "headers": {
+      "type": "object",
+      "properties": {
+        "properties": {
+          "type": "object"
+        }
+      },
+      "required": ["properties"]
+    },
```
magicmatatjahu, line 14: In places when you have `Schema Object` type you should write reference to JSON Schema spec like:

```json
"query": {
  "$ref": "http://json-schema.org/draft-07/schema#"
}
```

Or use our AsyncAPI Schema -> https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#L359

magicmatatjahu (commented):
@KhudaDad414 I added some comments :) Good progress!

KhudaDad414:
@derberg @magicmatatjahu The schemas are completed.

GeraldLoeffler:
I suggest to:
- use JSON Schema draft-07, just like the AsyncAPI spec
- give every schema a unique ID, maybe following the pattern `"$id": "http://asyncapi.com/<protocol>/bindings/server.json"`
- use `"additionalProperties": true` throughout for forwards compatibility

derberg:
I agree with @GeraldLoeffler with all except the `additionalProperties`, it must be `false` and only extensions should be allowed (as with `patternProperties`) like it is with the main spec. 

One more comment from my side: `definition.specificationExtension` is repeated everywhere, I'm wondering if it can be moved to a single fine in this repo and referenced everywhere, or we should refer this definition from the spec file. I don't have strong opinion here, thoughts? 

@rcoppen would you like to review schemas provided for IBMMQ?

GeraldLoeffler:
fwiw, here is the section from my work-in-progress bindings specification that argues for `additionalProperties: true`: https://github.com/integrational/asyncapi-bindings/tree/master/anypointmq#backwards-compatibility

derberg:
@GeraldLoeffler I looked at it and

> All bindings defined in this specification allow additional, unspecified fields, so that a concrete instance of a binding object MAY contain arbitrary fields not defined in this specification

This is exactly the purpose of extensions.

> As an example of backwards compatibility, consider a concrete binding object valid against version 1.1.0 of this specification. This binding object may specify bindingVersion: 1.1.0. This binding object may contain fields specific to version 1.1.0, which were not yet defined in, say, version 1.0.5 of this specification. However, because additional fields are allowed, this binding object is also valid against version 1.0.5 of this specification. The binding object valid against version 1.1.0 can therefore be used in a system that is only aware of version 1.0.5, although all fields not yet defined in version 1.0.5 of this specification will be ignored.

this is somehow valid, but did you consider that someone can add a property that means "A" and with the next binding version we add another property like that, but with a different meaning, "B"? and it can lead to a problem. This is why extensions are always with prefix

KhudaDad414:
@derberg made the recommended changes and used _https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/schema_ instead of _http://json-schema.org/draft-07/schema_ inside specification.

amqp/json_schemas/channel.json
```diff
@@ -0,0 +1,68 @@
...
+  "$id": "http://asyncapi.com/amqp/bindings/channel.json",
+  "title": "Channel Schema",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
```
derberg, line 12: in spec for amqp you can find

> Defines what type of channel is it. Can be either queue or routingKey (default).
> When is=routingKey, this object defines the exchange properties.
> When is=queue, this object defines the queue properties.

So you can either have `routingKey` or `queue`, but not both. Yes, first of all looks like example is wrong.
You need to therefore modify the json schema a bit as you need to use `oneOf` to describe that only one of the objects is allowes, either `exchange` or `queue` depending on the `is` property.

It is another level of JSON Schema 😄 . Feel free to let me know if you need some help here. I worked on something similar recently [here](https://gist.github.com/derberg/4e419d6ff5870c7c3f5f443e8bd30535#file-asyncapi-websocket-kraken-yml-L259-L313) and [here](https://github.com/asyncapi/spec/pull/533/files#diff-8cbe9bd49981ca789c9540337203e60bc894cb6a1ce98315a4b5c30d5199d833R127-R176)

amqp/json_schemas/message.json
```diff
@@ -0,0 +1,26 @@
...
+    "contentEncoding": {
+      "type": "string"
+    },
+    "messageType": {
+      "type": "string",
+      "default": "latest"
+    },
+    "bindingVersion": {
+      "type": "string",
+      "default": "latest"
```
derberg, line 22: please fix as I suggested in another comment, 
it shoudl not be here and the same is for all the other `bindingVersion` props

amqp/json_schemas/message.json
```diff
@@ -0,0 +1,26 @@
...
+      "$ref": "https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+    "contentEncoding": {
+      "type": "string"
+    },
+    "messageType": {
+      "type": "string",
+      "default": "latest"
```
derberg, line 18: should be removed

amqp/json_schemas/operation.json
```diff
@@ -0,0 +1,57 @@
+{
```
derberg, line 1: oh just noticed something that will be difficult to present with schema, like that for example `mandatory` is allowed only if the operation of the binding is `publish` 😞

http/json_schemas/operation.json
```diff
@@ -0,0 +1,34 @@
...
+    "type": {
+      "type": "string",
+      "enum": ["request","response"]
+    },
+    "method": {
+      "type": "string",
+      "enum": ["GET", "PUT", "POST", "PATCH", "DELETE", "HEAD", "OPTIONS", "CONNECT", "TRACE"]
+    },
+    "query": {
+      "$ref": "https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/schema"
```
derberg, line 22: as in other comment, please do not point to master

http/json_schemas/operation.json
```diff
@@ -0,0 +1,34 @@
...
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+    "type": {
+      "type": "string",
+      "enum": ["request","response"]
+    },
+    "method": {
```
derberg, line 17: allowed only if `type=request`, please express it with schema

ibmmq/json_schemas/channel.json
```diff
@@ -0,0 +1,61 @@
...
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+    "destinationType": {
+      "type": "string",
+      "enum": ["topic", "queue"]
```
derberg, line 15: speficy default as `topic` please

ibmmq/json_schemas/channel.json
```diff
@@ -0,0 +1,61 @@
...
+    },
+    "queue": {
+      "type": "object",
+      "properties": {
+        "objectName": {
```
derberg, line 20: please specify it is required

ibmmq/json_schemas/channel.json
```diff
@@ -0,0 +1,61 @@
...
+          "type": "string",
+          "maxLength": 48
+        },
+        "isPartitioned": {
+          "type": "boolean"
```
derberg, line 25: please specify `default` as `false`

ibmmq/json_schemas/channel.json
```diff
@@ -0,0 +1,61 @@
...
+        },
+        "exclusive": {
+          "type": "boolean"
```
derberg, line 28: please specify `default` as `false`

ibmmq/json_schemas/channel.json
```diff
@@ -0,0 +1,61 @@
...
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+    "destinationType": {
+      "type": "string",
+      "enum": ["topic", "queue"]
+    },
+    "queue": {
```
derberg, line 17: can be present only if `destinationType=queue` so you need to use `oneOf`/`allOf` again

ibmmq/json_schemas/channel.json
```diff
@@ -0,0 +1,61 @@
...
+        "string": {
+          "type": "string",
+          "maxLength": 10240
+        },
+        "objectName": {
+          "type": "string",
+          "maxLength": 48
+        },
+        "durablePermitted": {
+          "type": "boolean"
```
derberg, line 44: set default please

ibmmq/json_schemas/channel.json
```diff
@@ -0,0 +1,61 @@
...
+        },
+        "lastMsgRetained": {
+          "type": "boolean"
```
derberg, line 47: set default please

ibmmq/json_schemas/message.json
```diff
@@ -0,0 +1,33 @@
...
+      "$ref": "https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+    "type": {
+      "type": "string",
+      "enum": ["string", "jms", "binary"]
+    },
+    "headers": {
+      "type": "string"
```
derberg, line 18: can be present only if `type=string` or `type=jms`

ibmmq/json_schemas/message.json
```diff
@@ -0,0 +1,33 @@
...
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+    "type": {
+      "type": "string",
+      "enum": ["string", "jms", "binary"]
+    },
+    "headers": {
```
derberg, line 17: is required only if `type=binary`, tbh I don't know if you can do this with json schema, probably by using the `if` expression

ibmmq/json_schemas/message.json
```diff
@@ -0,0 +1,33 @@
...
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "expiry": {
+      "type": "integer",
```
derberg, line 24: provide default please

derberg (changes requested):
Oh man this was a tremendous effort from your side, hats off 🙇🏼 
Sorry for delayed review but I had some bank holidays here in Poland.

I left few comments, sorry but mainly to improve schemas. I did `amqp`, `http` and `ibmmq` only as I noticed that things that needs improvement are mostly around the same, `oneOf` or `default`

Also please add `description` property and `example` property to all the properties from all the schemas. On the other hand it is not a hard requirement for me for a 1st version, you can do it in another PR if you do not want to overcomplicate this one.

For next review round, please, when you push fixes, don't do it randomly or in one huge commit. Would be awesome if you could do commit per fix. I mean, when you remove `"default": "latest"` from all the `bindingVersion`, then just do this change in all files and push as single commit. And then the rest

derberg:
one more thing, for $id I think it is better to do `http://asyncapi.com/bindings/amqp/channel.json` instead of `http://asyncapi.com/amqp/bindings/channel.json`

ibmmq/json_schemas/message.json
KhudaDad414: @derberg  I think you mean the opposit, in readme file it states that:
> headers MUST NOT be specified if type = string or jms

amqp/json_schemas/channel.json
```diff
@@ -0,0 +1,68 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://asyncapi.com/amqp/bindings/channel.json",
+  "title": "Channel Schema",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#/definitions/specificationExtension"
```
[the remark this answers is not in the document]
KhudaDad414: @derberg I don't exactly get it. should I change it to: https://raw.githubusercontent.com/asyncapi/asyncapi-node/master/schemas/2.0.0.json#/definitions/specificationExtension instead?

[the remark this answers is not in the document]
derberg: yup, exactly, this way there is no risk that some change in master would break someting

ibmmq/json_schemas/message.json
derberg: yup, sorry 😅

KhudaDad414:
@derberg  Sorry, I think I wasted too much of your valuable time with this PR. I think I made all of the requested changes.

ibmmq/json_schemas/channel.json
derberg: ☝🏼

derberg:
@KhudaDad414 yup, I was using https://www.jsonschemavalidator.net/ and yeah, I see it works now 🤔 I must have done something wrong judging to quickly, sorry 🙏🏼 

> and about bindingVersion I don't exactly know what to do. should I use the version that is inside README.md file as the default value?

yeap, have a look at asyncapi spec for example https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#L17-L22

derberg:
sorry, wrong button 😅

KhudaDad414:
@derberg A review would be appreciated :)

amqp/json_schemas/channel.json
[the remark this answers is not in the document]
magicmatatjahu: Or maybe copy and paste `specificationExtension` to separate `common_bindings.yaml` file?

amqp/json_schemas/operation.json
magicmatatjahu: It's possible to do that by `if/then/else` but we must infer type from specification schema, so only way to do that is probably custom validator in parser.

magicmatatjahu:
@KhudaDad414 Related to oneOf and `not: required: []`, you can use false schema like (for example amqp):

```js
  "oneOf": [
    {
      "properties": {
        "is": { "const": "routingKey" },
        "queue": false
      },
      "required": [
        "exchange"
      ]
    },
    {
      "properties": {
        "is": { "const": "queue" },
        "exchange": false
      },
      "required": [
        "queue"
      ]
    }
  ],
```

But using `not: required` is also ok :)

KhudaDad414:
> @KhudaDad414 Related to oneOf and `not: required: []`, you can use false schema like (for example amqp):
> 
> ```js
>   "oneOf": [
>     {
>       "properties": {
>         "is": { "const": "routingKey" },
>         "queue": false
>       },
>       "required": [
>         "exchange"
>       ]
>     },
>     {
>       "properties": {
>         "is": { "const": "queue" },
>         "exchange": false
>       },
>       "required": [
>         "queue"
>       ]
>     }
>   ],
> ```
> 
> But using `not: required` is also ok :)

Thanks, didn't know that.🙂

amqp/json_schemas/operation.json
derberg: yeah, let's not overcomplicate, this is not a blocker for this PR for sure

derberg (approved):
@KhudaDad414 this is looking great now. This is a gigantic effort you did here! like seriously 🚀 

once we merge, we need to think on the next step, how to integrate it with `asyncapi-node` repo

magicmatatjahu (approved):

PR #65, opened 2021-06-15
fix: move WebSockets binding details to server and remove the method option

fmvilas:
**Description**

While it's true that WS doesn't have the concept of channels/topics/message-type, most people don't use a connection per message type but instead, they use a single connection and a subprotocol. Examples of subprotocols are Socket.io, STOMP over WS, custom ones, etc. For that reason, I'm moving all the WS options to the Server Binding Object.

Also, as per the [WebSocket RFC](https://datatracker.ietf.org/doc/html/rfc6455):

> The method of the request MUST be GET, and the HTTP version MUST be at least 1.1.

Therefore, I'm removing the `method` option.

### DIFF
--- a/websockets/README.md
+++ b/websockets/README.md
@@ -6,15 +6,22 @@ This document defines how to describe WebSockets-specific information on AsyncAP
 
 ## Version
 
-Current version is `0.1.0`.
+Current version is `0.2.0`.
 
 
 <a name="server"></a>
 
 ## Server Binding Object
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
+##### Fixed Fields
 
+Field Name | Type | Description
+---|:---:|---
+<a name="operationBindingObjectQuery"></a>`query` | [Schema Object][schemaObject] | A Schema object containing the definitions for each query parameter to use when establishing the connection. This schema MUST be of type `object` and have a `properties` key.
+<a name="operationBindingObjectHeaders"></a>`headers` | [Schema Object][schemaObject] | A Schema object containing the definitions of the HTTP headers to use when establishing the connection. This schema MUST be of type `object` and have a `properties` key.
+<a name="operationBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
+
+This object MUST contain only the properties defined above.
 
 
 
@@ -22,18 +29,8 @@ This object MUST NOT contain any properties. Its name is reserved for future use
 
 ## Channel Binding Object
 
-When using WebSockets, the channel represents the connection. Unlike other protocols that support multiple virtual channels (topics, routing keys, etc.) per connection, WebSockets doesn't support virtual channels or, put it another way, there's only one channel and its characteristics are strongly related to the protocol used for the handshake, i.e., HTTP.
+This object MUST NOT contain any properties. Its name is reserved for future use.
 
-##### Fixed Fields
-
-Field Name | Type | Description
----|:---:|---
-<a name="operationBindingObjectMethod"></a>`method` | string | The HTTP method to use when establishing the connection. Its value MUST be either `GET` or `POST`.
-<a name="operationBindingObjectQuery"></a>`query` | [Schema Object][schemaObject] | A Schema object containing the definitions for each query parameter. This schema MUST be of type `object` and have a `properties` key.
-<a name="operationBindingObjectHeaders"></a>`headers` | [Schema Object][schemaObject] | A Schema object containing the definitions of the HTTP headers to use when establishing the connection. This schema MUST be of type `object` and have a `properties` key.
-<a name="operationBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
-
-This object MUST contain only the properties defined above.
 
 <a name="operation"></a>
 


smoya:
> most people don't use a connection per message type but instead, they use a single connection and a subprotocol.

With this change, do we keep giving support for those use cases where several WS endpoints are mounted in the same server? For example, I imagine a server with two WS endpoints: `/subscribe-events` and `/publish-events` where both headers and query params are different.

fmvilas:
> For example, I imagine a server with two WS endpoints: /subscribe-events and /publish-events where both headers and query params are different.

That's not possible with WebSockets. The `query` and `headers` are for the HTTP server not for the WS server. It's only for this initial HTTP Upgrade request. If you have 2 WS servers over a single HTTP server, it's exactly why you need to have this information on the Server Bindings Object. The way it's currently described in the binding means people would have to create an HTTP server per channel, and that's why I want to change it. That's possible but an overkill IMHO.

derberg:
binding would cover what technically is possible, and even though some these days use subprotocols, like in cryptocurrency trading API, this is not the case always.

also thing is that path to endpoint is the channel, so it is strange that query will be on the server and not channel level. Have a look at our official example -> https://github.com/asyncapi/spec/blob/master/examples/2.0.0/websocket-gemini.yml

technically you can also have 2 endpoints on the server https://github.com/derberg/shrekapp-asyncapi-designed/blob/main/asyncapi.yaml at the same time

smoya:
> It's only for this initial HTTP Upgrade request

That's the point. You still can send different Headers that affect the upgrade. As per my example, two endpoints (yes, HTTP that upgrade to WS). Isn't that the way you specify wich subprotocol to use (Header `Sec-WebSocket-Extensions` for example)?

fmvilas:
Ok so then I guess we need to cover all the cases. For instance, in the following HTML code:

```html
<script>
  const socket = new WebSocket('ws://localhost:3005')
  socket.addEventListener('message', (event) => {
    const message = JSON.parse(event.data)
    console.log(message.type, JSON.parse(message.data))
  })
  socket.addEventListener('open', () => {
    socket.send('{"type": "user/signedup", "data": {"displayName": "Fran Mendez", "email": "fmvilas@gmail.com"}}')
  })
</script>
```

I only have a single connection and multiple message types (channels). I can't specify different `query` params and `headers` for each message type because there's just a single connection and a single HTTP Upgrade request. Therefore, I want to have this information on the Server Binging Object not on the Channel Binding Object.

Should we then have the same information in Server Binding Object and Channel Binding Object and make the Channel Binding Object prevail over the Server Binding Object?

smoya:
> Should we then have the same information in Server Binding Object and Channel Binding Object and make the Channel Binding Object prevail over the Server Binding Object?

That makes sense to me. Even though it complicates the operative, but represents valid use cases.

derberg:
why in your implementation message translates into a channel? looks like something specific to socket.io and not WebSocket, or I'm missing something 🤔

fmvilas:
What I'm showing here is an example of a custom subprotocol. The `type` field is actually the channel and the `data` field is the actual message. Since WS doesn't support channels, an option is to have them encoded as part of the message. That's the most common way to use WS: Socket.io, WAMP, STOMP over WS, Pusher, and more.

derberg:
> What I'm showing here is an example of a custom subprotocol

you mean the `type` part is about custom subprotocol, right? Does it make sense to put it in the channel? Sounds like mixing one protocol with another. Have a look at https://github.com/asyncapi/spec/issues/457

fmvilas:
It does make sense to put it on the channel IF you have a connection per channel, and therefore you have the opportunity to define the subprotocol on connect. However, if you only have one connection and multiple channels (virtual ones, using something like `type` field or similar approach like Socket.io) then it does make sense to have it on the server instead of the channel.

It really depends on how you use WebSockets.

fmvilas:
> you mean the type part is about custom subprotocol, right?

Also, the whole message is structured using a custom subprotocol (it doesn't have a name). The real message payload is inside `message.payload.data`. The `type` part is just a construct for the application to understand what type of message is it or, in other words, what channel is it.

dedoussis:
As mentioned, a Socket.IO server may have multiple channels, which are usually virtual, but could be forced to be physical. See the `forceNew` option when initialising a connection: https://socket.io/docs/v4/client-initialization/#forceNew

Hence Socket.IO requires bindings on the channel level. In the coming weeks I will be publishing a blog post that thoroughly covers the topic.

fmvilas:
> a Socket.IO server may have multiple channels, which are usually virtual, but could be forced to be physical

But what if I don't want to force it? I mean, I don't want the spec to force a specific implementation. It's usually very convenient to have virtual channels (which is the default btw). In that case, we need the handshake details on the Server Binding instead.

I'm starting to think the best solution here is to have it in both places and give the Operation Binding precedence over the Server one.

dedoussis:
> But what if I don't want to force it? I mean, I don't want the spec to force a specific implementation. It's usually very convenient to have virtual channels (which is the default btw). In that case, we need the handshake details on the Server Binding instead.

The convention I propose in my blog post is the following:
Every Socket.IO server has a default namespace (`/`) which MAY have bindings. Since namespaces are by default virtual, there is no need to repeat the same bindings under each custom namespace (channel). Each custom namespace inherits the bindings of the default namespace.
The inclusion of bindings under a custom namespace signals the user to use the `forceNew` option when connecting to that namespace.

But I kind of agree with you that it wouldn't harm to have both server and channel bindings in the `ws` protocol.
Although one could also claim that `socketio` should be a protocol of its own as it could be implemented solely using HTTP polling.

derberg:
Shouldn't actually `socketio` have a dedicated binding? https://github.com/socketio/socket.io-protocol

Throwing a bomb into the discussion 😄

fmvilas:
> Although one could also claim that socketio should be a protocol of its own as it could be implemented solely using HTTP polling but let's not derail this discussion.

I actually think that's the key. It's a subprotocol of two different protocols (HTTP and WS). And I don't think it's derailing the discussion but actually making it better 😊 Would love to think how this would play with something like this: https://github.com/asyncapi/spec/issues/457.

> Shouldn't actually socketio have a dedicated binding? https://github.com/socketio/socket.io-protocol
>
> Throwing a bomb into the discussion 😄

💣 but actually right and related to what we discussed above. If we end up considering Socket.IO a protocol on its own, it should have its own binding too.

derberg:
@gexclaude you might be interested with this discussion

fmvilas:
Yup, thanks for the pointer @smoya. It's indeed relevant.

derberg:
looking at https://github.com/asyncapi/bindings/pull/144
shouldn't `query` stay in channel and `headers` move to server?

in websocket header is for authorization, when you connect, so header should be described together with authorization under server 🤷🏼

fmvilas:
Depends on how you understand the relationship between channels and WS. For instance, WS doesn't support channels/topics so we may decide to create a different connection for each channel. In this case, headers (and authorization too) may be different per channel. However, we can create multiple server definitions with different authorization mechanisms (or different headers altogether). I don't have a strong opinion on where it should live.

derberg:
I'm supporting the approach where each channel is a separate connection

> However, we can create multiple server definitions with different authorization mechanisms (or different headers altogether)

exactly ☝🏼 
I mean, we have mechanisms that can help do it this way. So we keep things consistent.
or we need to enable `security` reference on channel level as we did with operation. But I do not think so. Technically, in WS you do not connect to server first, and then channel, but basically combine server URL with channel name and connect once

fmvilas:
Yes, you're right. I think we should be consistent and do it this way. Otherwise, we have to add security on the channel and that's even worse 😅

Issue #68, opened 2021-06-24
Understanding WebSocket API description's query and Chanels Object keys


bali182:
Hello, I'm raising this issue  to have a better understanding how the ws protocol is described. I'm looking at this example: 

https://github.com/asyncapi/spec/blob/master/examples/2.0.0/websocket-gemini.yml

1. When using websockets, the keys in `Channels Object` are the URLs to connect to right?  So this:
```yaml
channels:
  /v1/marketdata/{symbol}:
```

Would mean connect to this URI (where symbol is a parameter), right?:
```ts
//                  in servers              in the channel
new WebSocket("wss://api.gemini.com" + "/v1/marketdata/{symbol}")
```

If we have multiple entries in the `Channels Object` does that mean we can connect to multiple endpoints with their respective parameters and available subscriptions/publishes?

2. What is query under ws bindings? Is that something that has to be turned into a query string when establishing the connection? So this:

```yaml
query:
  type: object
  properties:
    heartbeat:
      type: boolean
      default: false
    top_of_book:
      type: boolean
      default: false
    bids:
      type: boolean
      default: true
    ...
```
Would also go into the URI?
```ts
new WebSocket("wss://api.gemini.com" + "/v1/marketdata/{symbol}?heartbeat=...&top_of_book=...")
```
In case yes, how to serialize the fields of this object? I saw that it must be an object with properties in the docs, but can the properties objects themselves? If so how deeply nested can this structure go and what would be the serialization method for it? OpenAPI has a nightmarish amount of parameter serialization methods, I hope you went with a simpler approach :D

In case I'm completely wrong please advise!

bali182:
Cool thank you for moving it!

> I mean, how would the resulting query param look like for such example?

That is exactly my question :)  JSON Schema allows it. AsyncAPI docs don't say as far as I have seen that you can't do this.  So my question is:

A) Is nesting not allowed, and this info missing from the docs?
B) If nesting is allowed are there specs on serialization?

Assume we have the schema you mentioned:

```yaml
query:
  type: object
  properties:
    heartbeat:
      type: object
      properties:
        foo:
          type: string
        bar:
          type: number
```

OpenAPI would prescribe serializition of the equivalent like:
style: form, explode: true -> `?foo=someString&bar=1`
style: form, explode: false -> `?heartbeat=foo,someString,bar,1`
style: deepObject, explode: true -> `?heartbeat[foo]=someString&heartbeat[bar]=1`

[...]

Issue #66, opened 2021-06-16
Clarify schema restriction on WebSocket Bindings query


fujiapple852:
The documentation for [WebSockets Bindings](https://github.com/asyncapi/bindings/blob/master/websockets/README.md) states that the `query' must be:

> A Schema object containing the definitions for each query parameter. This schema MUST be of type object and have a properties key.

I am attempting to encode an existing WebSocket API which establishes the connection with a POST request which has a body that is a JSON array. This appears to be accepted by several validators (I can produce an example if required) yet is not allowed by the specification.

Is the restriction as documented accurate?

derberg:
@fujiapple852 it would be great to get an example. You can rework this one.

Also keep in mind that validators do not support bindings yet. So whatever goes in is treated as valid

fujiapple852:
@derberg Apologies for the delay and thank you for the assistance.

Here is an example of a trivial `bindings` object which specifies that a `POST` operation with a `query` (which I interpret as the body of the `POST` as there is no separate `body` field specified) is an `array` of `string`:

```json
{
  "ws": {
    "method": "POST",
    "query": {
      "type": "array",
      "items": {
        "type": "string"
      }
    },
    "bindingVersion": "0.1.0"
  }
}
```

Here is the sample example embedded in a minimal valid AsyncAPI document:

```json
{
  "asyncapi": "2.0.0",
  "info": {
    "title": "Test",
    "version": "1.0.0"
  },
  "channels": {
    "myChannel": {
      "subscribe": {
        "operationId": "myOperation",
        "message": {
          "contentType": "application/json",
          "payload": {
            "$ref": "#/components/schemas/Events"
          }
        }
      },
      "bindings": {
        "ws": {
          "method": "POST",
          "query": {
            "type": "array",
            "items": {
              "type": "string"
            }
          },
          "bindingVersion": "0.1.0"
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Events": {}
    }
  }
}
```

The example appears to violate the specification because the `schema` object has the type `array` rather than the type required type `object`.

As you point out, the validators do not support bindings yet and so I would like to confirm if the example above is intended to be valid in the specification (and if not what the correct alternative representation should be)?

derberg:
> You can rework this one

of course I did not provide a link to an example, 🤦🏼 , sorry

[...]

PR #69, opened 2021-07-04
docs: clarifying websocket params

bali182:
**Description**

- Clarifies that query parameters must be primitive.
 
**Related issue(s)**
#68 #66

### DIFF
--- a/websockets/README.md
+++ b/websockets/README.md
@@ -29,8 +29,8 @@ When using WebSockets, the channel represents the connection. Unlike other proto
 Field Name | Type | Description
 ---|:---:|---
 <a name="operationBindingObjectMethod"></a>`method` | string | The HTTP method to use when establishing the connection. Its value MUST be either `GET` or `POST`.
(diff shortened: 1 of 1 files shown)


KhudaDad414:
Hey @bali182, can you add your changes to the [channel json schemas] (https://github.com/asyncapi/bindings/blob/master/websockets/json_schemas/channel.json#L22-L29) as well?

thushalya:
hello derberg  @derberg , is this valid for version 2.5.0 also?   

> We already had a sentence that it must be an object with properties, we just need to add that properties of the object cannot be of type object and array

there is no statement here https://github.com/asyncapi/bindings/tree/master/websockets#:~:text=A%20Schema%20object%20containing%20the%20definitions%20for%20each%20query%20parameter.%20This%20schema%20MUST%20be%20of%20type%20object%20and%20have%20a%20properties%20key.

It may be because this pull request hasn't yet been merged.

Isn't it good if we can add that statement or define a good parameter serialization method like in OpenAPI spec? 
https://swagger.io/docs/specification/serialization/
Need a quick reply

PR #70, opened 2021-07-09
[docs:] Add conventions for bindings

### LINKED ISSUE #62: Establish conventions on where we define configuration for a protocol using bindings for consistency
# Documenting Producers and Consumers

## Proposal

We should agree conventions around bindings so that different protocols are as consistent as possible as to how we use bindings on objects that support them. Today, implementers of protocols have not been consistent in application of bindings. This makes it difficult to create tooling that works with different bindings.

## Discussion

Let's assume that I want to document the API exposed by a microservice, Foo, and I want to consume it from a microservice, Bar, that wants to receive via RMQ (AMQP 0-9-1), and another microservies, Wibble, that wants to receive receive via Kafka. 

It is a little contrived, but the scenario let's us do a comparison of binding needs.

For reference then:

||Producer|Consumer|
|-----|:------:|:-------:|
|AMQP 0-9-1|Foo|Bar|
|Kafka|Foo|Wibble|

Now let's assume we have a channel quux which we communicate via. Foo sends a message to quux and Bar and Wibble receive a message from quux.

How would we use AsyncAPI to document this setup?

First, I would assume that we have an AsyncAPI files for the producer, Foo, and separate files  for the consumers, Bar and Wibble.

Let's assume we define the Info and Server properties for both, and that the Server properties for both include a range of environments. Of course, I might use $ref to avoid a repeated definition of the object, and instead refer to a common definition.

Instead let's think about the channel. 

```yml
channels:
    quux:
        subscribe:
            summary: Messages about quux
            operationId: sendQuux
            $ref: "#/components/messages/quuxEvent"
```

On the producer, I want to define the channel and a subscribe operation on it - to indicate that Foo produces this topic. As the producer, I need to set up this channel. What do I need to define so that I could use the Generator to create Infrastructure as Code (IAC) such as Terraform templates or an Ansible playbook to be able to send messages to this channel.

### AMQP 0-9-1

For RMQ I need an exchange, this governs how messages are routed. I also need a routing key for this channel. We can assume that the channel name is the routing key name. But we need to define the exchange, so I need a binding. From the amqp binding, I'll pull the binding for an exchange. (see https://github.com/asyncapi/bindings/tree/master/amqp)

Logically, where should this be defined? Does it belong to the channel or the subscribe operation on the channel?

It's not uncommon to have a pattern where a consumer creates the channel and binds its queue to it, ahead of a producer sending to it. So we might also want to define the exchange on the consumer AsyncAPI file. It makes sense then that we do not define this on the subscribe operation, as we will not have that on the consumer API file, and adding it to the consumer API file will just increase the confusion around publish-consumer, subcribe-producer.

So we can make an initial assumption: *by convention, bindings to set up the channel we communicate via live at the channel level as they may appear on both sides.*

Let's look at Foo's AsyncAPI file:

```yml
channels:
    quux:
        subscribe:
            summary: Messages about quux
            operationId: sendQuux
            $ref: "#/components/messages/quuxEvent"
    bindings:
        amqp:
            exchange:
                name: myExchange
                type: topic
                durable: true
                autoDelete: false
                vhost: /
                bindingVersion: 0.1.0
```

Now, https://github.com/asyncapi/bindings/tree/master/amqp defines the queue here as well.

But, a queue belongs to a consumer, not a producer. I may have many consuming services, each will have their own queue if they want to obtain their own copy of the message. Logically then, I don't want my producer's AsyncAPI file to contain the definition of my consumers. The pub-sub model desires to prevent a producer having to know its consumers, so as to avoid that coupling. (Of course we may want to take the information from both sides and build a graph to document the flow).

I don't want to define a consumer's queue ijn the producer's file but in the consumer then. So where should that binding go. The channel or the operation? I would argue that the consumption belongs with the operation, because we don't possess a queue without that. If I add it to the channel I also risk confusing users of AsyncAPI. If I define a queue against the channel should it appear in both producer and consumer documents? If I define it against the publish operation, this becomes clearer.

So we can make a second assumption: *by convention, bindings to configure how we consume a channel live with the publish operation on the consumer*

Let's look at what Bar's AsyncAPI file looks like if we do this (note that we show the exchange definition again, in case we opt for consumer creates, and of course a $ref could optimize this):

```yml
channels:
    quux:
        publish:
            summary: Messages about quux
            operationId: sendQuux
            $ref: "#/components/messages/quuxEvent"
            bindings:
                amqp:
                    queue:
                        name: myQueue
                        type: topic
                        durable: true
                        autoDelete: false
                        bindingVersion: 0.1.0
    bindings:
        amqp:
            exchange:
                name: myExchange
                type: topic
                durable: true
                autoDelete: false
                vhost: /
                bindingVersion: 0.1.0
```

Finally, there are some things that impact how the producer operates that are independent of the channel. If we want to generate the producer code, we would need to configure these for the subscribe operation and not for the channel, as they make no sense on the consumer side, as we won't be generating producer code there. The main one for RMQ is enabling publisher confirms.

So we can make a third assumption: *by convention, bindings to configure how we produce to a channel live with the subscribe operation on the producer*

If we add something that does this (not in the existing specification so pseudo), our final producer file looks more like

```yml
channels:
    quux:
        subscribe:
            summary: Messages about quux
            operationId: sendQuux
            $ref: "#/components/messages/quuxEvent"
            bindings:
                amqp:
                    producer:
                        confirmSelect: true
    bindings:
        amqp:
            exchange:
                name: myExchange
                type: topic
                durable: true
                autoDelete: false
                vhost: /
                bindingVersion: 0.1.0
```

### KAFKA

Kafka does not have an exchange. It's routing is uncomplicated, just subscribe to the topic. 

But, for IAC there are properties we might want to configure for a given channel that do not appear in the existing specification. We'll add a couple of examples so we can see how it would look. The Admin client gives some clues as to what is needed here: https://kafka.apache.org/23/javadoc/index.html?org/apache/kafka/clients/admin/AdminClient.html

We already have amqp bindings against the channel and by following the convention above, these will be defined at the same level, which makes it easy to reason about, and write a generator for.

This leads to a meta-convention: *to facilitate multiple-protocols binding authors should strive to use equivalent binding targets for their protocol, where possible*

The flexibility of AsyncAPI is powerful, but if we are to create tooling we need some predictability for protocols that can conform.

In addition, how we send messages the topic has a lot of options as to how we consume. Now a lot of those may be client-specific. So the Confluent client does not necessarily have the same set of properties as other client SDKs for Kafka, and languages may vary. This is an interesting question because we may wonder if the binding here is not Kafka but Confluent? I'm not going to resolve that one in this issue, but it's worth bearing in mind for folks working on a binding.

Anyway, following our convention, if it affects how we would generate code for a producer, it ought to bind to the subscribe operation:

Here is Foo's AsyncAPI file now we have added those bindings:

```yml
channels:
    quux:
        subscribe:
            summary: Messages about quux
            operationId: sendQuux
            $ref: "#/components/messages/quuxEvent"
            bindings:
                amqp:
                    producer:
                        confirmSelect: true
                kafka:
                    producer:
                        transactionalId: Foo
                        replication: Acks.All
                        batchMessages: 10
                        retries: 3
                        queue_strategy: fifo
                        max_in_flight: 5
                     
    bindings:
        amqp:
            exchange:
                name: myExchange
                type: topic
                durable: true
                autoDelete: false
                vhost: /
                bindingVersion: 0.1.0
        kafka:
            topic:
                partitioner: ConsistentRandom
                numOfPartitions: 6
                replicationFactor: 3
```

How about the consumer? Of course Wibble has it's own file. 

What are the key properties we need to create a client? 

I would argue for the consumer group. Whilst we might consider this an implementation detail, having the group defined at the level of the AsyncAPI binding may help us with understanding what consumer a group relates to in any telemetry. We might make the same argument for consumer ids, but I would not tend to find how many consumers there were in our group in an AsyncAPI file. 

There are other properties that we might want for code generation though such as those that govern how we deal with offsets. Again, there is an open question if the binding here is Kafka, or Confluent for the client SDK. I'll park that as well, as I don't think it changes the overall discussion either way it ought to be bound to the publish operation on the channel.

When we follow the publish route as above Wibble's file looks like this (it's less common to have the consumer create the topic than with RMQ, but let's show that anyway for completeness):

```yml
channels:
    quux:
        publish:
            summary: Messages about quux
            operationId: sendQuux
            $ref: "#/components/messages/quuxEvent"
            bindings:
                kafka:
                    consumer:
                        group: myGroup
                        maxPollIntervalInMs: 100
                        enableAutoOffsetStore: true
                        enableAutoCommit: true
    bindings:
        kafka:
            topic:
                partitioner: ConsistentRandom
                numOfPartitions: 6
                replicationFactor: 3
```

## Conclusion

By contrast to the *laissez-faire* approach to bindings today, establishing some conventions would help us with consistency around bindings and the ability to create tooling around definitions. Whilst we may not desire the conventions above, having some agreement about where we define some of the binding properties would help both tool authors and the authors of future bindings, who would have a stronger understanding of how to implement thier own binding

iancooper:
**Description**
This establishes a set of conventions to use when designing bindings. The goal is that bindings should be consistent, such that they can be easily interpreted by tools. 'See #62` for an outline of this problem.

The conventions are listed as a series of items. The intent here is that items could be added (or retired) by the community as we discover better ways to work.

I don't expect that everyone will agree with these conventions, but I do believe that the points here are worthy of community agreement on a convention. I don't believe that this list is complete, but I do believe it is a "good start" for discussion around the conventions we need. 

**Related issue(s)**
'Resolves #62`

### DIFF
--- a/.github/workflows/if-nodejs-version-bump.yml
+++ b/.github/workflows/if-nodejs-version-bump.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   version_bump:
-    name: Generate assets and bump
+    name: Bump version
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
@@ -26,9 +26,6 @@ jobs:
       - if: steps.packagejson.outputs.exists == 'true'
         name: Install dependencies
         run: npm install
-      - if: steps.packagejson.outputs.exists == 'true'
-        name: Assets generation
-        run: npm run generate:assets
       - if: steps.packagejson.outputs.exists == 'true'  
         name: Bump version in package.json
         # There is no need to substract "v" from the tag as version script handles it


Conventions.md
```diff
@@ -0,0 +1,301 @@
...
+## Why Adopt These?
+AsyncAPI does not constrain how Bindings are defined, beyond a valid JSON schema.  This makes Bindings flexible extension points for implementers. This flexibility is powerful, but it comes at a price in that it can be hard for implementers to understand how to use Bindings consistently, which makes it difficult for those writing tools that work with AsyncAPI to support a range of protocols.
+
+## What Belongs on a Binding?
+The Binding should be used to provide configuration data that a protocol requires. That configuration may be exposed via Message-Oriented-Middleware (MoM) using a Management API that allows code to configure the MoM as described. The fields in a Binding should represent protocol specific metadata, that is not exposed as an existing field on the Channel, Operation or Message objects.
+
+A Binding may also be used by the providers of SDKs for MoM to provide metadata to configure producers or consumers using that SDK. That is outside the scope of advice here, but SDK owners wishing to use Bindings may follow the general advice here.
+
+## Effective Bindings
+### **Item 1** Use the Extensions Format
```
fmvilas, line 14: This is still experimental. I'd not recommend this yet. Maybe we should just advise using Specification Extensions: https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#specificationExtensions.

Conventions.md
```diff
@@ -0,0 +1,301 @@
...
+Prefer use of the [Bindings Extensions](https://github.com/asyncapi/extensions-catalog) format when defining a binding over raw JSON Schema.
+
+The Binding Extensions format works with tools because it indicates where in the specification the binding should be used via a Hook. Using a Hook signals clearly to the user of an extension on which of Channel, Operation or Message the binding can be used. Implementers may choose to use author a binding so that the only fields on the binding are those appropriate to the object identified by the Hook. It also makes it possible for tools to determine correctness of an AsyncAPI file by ensuring that the binding is only used where permitted.
```
fmvilas, line 17: I know I've said we should not advise this yet but just in case we end up leaving this item here:

```suggestion
The Binding Extensions format works with tools because it indicates where in the specification the binding should be used via a Hook. Using a Hook signals clearly to the user of an extension on which of Server, Channel, Operation or Message the binding can be used. Implementers may choose to use author a binding so that the only fields on the binding are those appropriate to the object identified by the Hook. It also makes it possible for tools to determine correctness of an AsyncAPI file by ensuring that the binding is only used where permitted.
```

Conventions.md
```diff
@@ -0,0 +1,301 @@
...
+
+Not all uses of a Binding have Infrastructure as Code as a goal, and not all parameters to API calls tend to be required, so it is important to avoid using required fields on any bindings where the information is likely needed in IaC or Contract-First scenarios [see Item 4](#item-4-support-contract-first--scaffolding).
+
+Some transports do not provide a specification of a management API used to create infrastructure but instead have middleware specific APIs. To support a protocol that does not define infrastructure creation as part of its standard, consider a binding for the middleware in addition to the protocol i.e. AMQP 1-0-0.
+
+### **Item 4** Support Contract First && Scaffolding
+Design your binding to support Contract First definition of an endpoint that is later used with [Scaffolding](https://en.wikipedia.org/wiki/Scaffold_(programming)) to generate the skeleton of an application. Contract Fist scenarios requires the binding to support the necessary parameters to either produce or consume messages via the API or SDK.
```
fmvilas, line 38: ⚖️ ✊ 😛 

```suggestion
Design your binding to support Contract First definition of an endpoint that is later used with [Scaffolding](https://en.wikipedia.org/wiki/Scaffold_(programming)) to generate the skeleton of an application. Contract-First scenarios require the binding to support the necessary parameters to either produce or consume messages via the API or SDK.
```

Conventions.md
```diff
@@ -0,0 +1,301 @@
...
+
+Code generation from an AsyncAPI definition, particularly for consumers, is a use case that improves developer productivity and tends to drive adoption. Not all uses of a Binding have later generation of code via Scaffolding as a goal. 
+
+There may be multiple SDKs for a particular protocol, for different languages and frameworks. In many cases the variables required by these targets remain common across the targets and can simply be inferred from the protocol. In this case it is possible to define a common set of fields for a binding that can be used by Scaffolding regardless of the SDK. In other cases, SDKs vary considerably for the middleware. In this case you may prefer to provide a binding for the middleware that facilitates code generation for that middleware.
```
fmvilas, line 42:
```suggestion
There may be multiple SDKs for a particular protocol, for different languages and frameworks. In many cases, the variables required by these targets remain common across the targets and can simply be inferred from the protocol. In this case, it is possible to define a common set of fields for a binding that can be used by Scaffolding regardless of the SDK. In other cases, SDKs vary considerably for the middleware. In this case, you may prefer to provide a binding that facilitates code generation for a specific SDK.
```

Conventions.md
```diff
@@ -0,0 +1,301 @@
...
+
+Where a protocol defines metadata that forms part of the message sent via middleware it should be defined as part of the header schema (potentially as MessageTraits), and where the protocol defines data (or transmits metadata in the payload) it should be defined as part of the payload. Note that the payload does not support MessageTraits. 
+
+Tooling to validate that messages match the schema will expect to verify against the schema of the message payload and headers, putting message metadata into a Message Binding makes this harder as tooling must know to look there as well. 
+
+Implementers of producers or consumers are also likely to assume that the the can refer to #/Components/Schema to understand payloads and headers, and adding metadata into bindings in addition makes it harder to comprehend the schema of a message.
```
fmvilas, line 198:
```suggestion
Implementers of producers or consumers are also likely to assume that they can refer to #/Components/Schema to understand payloads and headers, and adding metadata into bindings, in addition, makes it harder to comprehend the schema of a message.
```

Conventions.md
```diff
@@ -0,0 +1,301 @@
...
+...
+
+components:
+    schemas:
+       quux:
+	 type: object
+	 properties:
+	     quuxDetails: 
+	     	type: string	 
+       amqpBasicProperties:
```
fmvilas, line 219: What is this?

fmvilas (commented):
I love the direction this is taking, @iancooper. I think some things related to `publish` meaning "we receive" and `subscribe` meaning "we send", will be solved soon ™️ 

Have a look at this video where Lorna and I discuss potential solutions for the publish/subscribe challenge: https://www.youtube.com/watch?v=YixYuYCmyJs.

Conventions.md
iancooper: I think it would be useful to adopt something that allowed us to provide a consistent format for a binding, other than just JSON, particularly around tooling.

damaru-inc:
I really like this.

iancooper:
Apologies all, wei/pull seemed to kill the PR when it updated my fork of bindings. I didn't appreciate that it does not seem to be playing nicely with a PR from master on my fork.

I'll close this PR and use #75 

@fmvilas I'll see if I can figure out how to pick up your valuable changes and re-apply them

iancooper:
All I have raised a new PR that should not break when wei/pull syncs my fork, updated for @fmvilas comments. So I will close this PR, please use the other one. Apologies

PR #81, opened 2021-09-13
feat(kafka): add AVRO type in message key binding

M3lkior:
Refer to asyncapi/avro-schema-parser#67

Here is the related update due to the support of avro file in kafka key bindings

### DIFF
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -6,7 +6,7 @@ This document defines how to describe Kafka-specific information on AsyncAPI.
 
 ## Version
 
-Current version is `0.1.0`.
+Current version is `0.2.0`.
 
 
 <a name="server"></a>
@@ -69,7 +69,7 @@ This object contains information about the message representation in Kafka.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="messageBindingObjectKey"></a>`key` | [Schema Object][schemaObject] | The message key.
+<a name="messageBindingObjectKey"></a>`key` | [Schema Object][schemaObject] / [AVRO file](https://github.com/asyncapi/avro-schema-parser) | The message key.
 <a name="messageBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
 
 This object MUST contain only the properties defined above.
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -69,7 +69,7 @@ This object contains information about the message representation in Kafka.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="messageBindingObjectKey"></a>`key` | [Schema Object][schemaObject] / [AVRO file](https://github.com/asyncapi/avro-schema-parser) | The message key.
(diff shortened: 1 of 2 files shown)


kafka/README.md
```diff
@@ -69,7 +69,7 @@ This object contains information about the message representation in Kafka.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="messageBindingObjectKey"></a>`key` | [Schema Object][schemaObject] | The message key.
+<a name="messageBindingObjectKey"></a>`key` | [Schema Object][schemaObject] / [AVRO file](https://github.com/asyncapi/avro-schema-parser) | The message key.
```
fmvilas, line 72: 🤔 I don't think `Avro file` is the allowed value here. This is saying that you can do this:

```yaml
key: 'path/to/user-create.avsc#/UserCreate'
```

Instead, what you can do here is provide a Reference Object that will point to an Avro file:

```yaml
key:
  $ref: 'path/to/user-create.avsc#/UserCreate'
```

I think you're going through a rabbit hole here: coming up with a way to reference other types of schemas. @magicmatatjahu already created an issue about that: https://github.com/asyncapi/spec/issues/622.

M3lkior: Hi @fmvilas ; i don't really understand what you are expected in the bindings' doc ; i update the `kafka/json_schemas/message.json ` with the expected value as `$ref` like 
```
key:
  $ref: 'path/to/user-create.avsc#/UserCreate'
```

For the new supported types, you suggest to point to the `$ref` way (https://asyncapi.io/docs/specifications/v2.1.0#referenceObject) ? With not specially an avro file format but any value in the `$ref` value ?

Concerning the @magicmatatjahu ; and due to the proposal DRAFT status, is it appropriate to document the new future way even if the implementation is not done ?

magicmatatjahu: As you wrote, my proposition is in a DRAFT/Strawman stage, so it might not be released until January or April next year (we release specs every 3-4 months). I would not change here the information that the `key` field is Schema Object or Avro file. I would just leave the information that it is Schema Object and add in the `**Note**` under the table that Schema Object for `key` can be represented with Avro format, something like that:

```md
> **NOTE**: `key` can be represented with AVRO format.
```

M3lkior: It seems right for me

fmvilas: Still my first comment applies. As it's defined right now you're not saying you can use `$ref`. You're saying you can do `key: path/to/avsc` and that's not what you want. You want:

```yaml
key:
  $ref: 'path/to/avsc'
```

However, given the way `$ref` works, it's going to be dereferenced into an Avro Schema object so I'd instead define it as:

```suggestion
<a name="messageBindingObjectKey"></a>`key` | [Schema Object][schemaObject] \| [AVRO Schema Object](https://avro.apache.org/docs/current/spec.html) | The message key.
```

Then, when resolving and dereferencing the `$ref`, it will become a valid definition and it also would allow you to inline the definition right there using Avro.

M3lkior: is the new proposal ok for you ?

fmvilas (approved):
Looks good to me 👍

M3lkior:
> LGTM



> LGTM, but we must remember about changing it in the future 😄

yes, but it allows to fix the regression from the nunjunks template to the react refactoring :)

magicmatatjahu:
@M3lkior Sorry 😅  What do you mean by:

> yes, but it allows to fix the regression from the nunjunks template to the react refactoring :)

?

PR #84, opened 2021-09-16
docs: proposed set of bindings for AWS - SNS and SQS

iancooper:
**Description**

- This is a proposed set of bindings for SNS and SQS
- They are complete enough to allow using AWS APIs to create infrastructure from the information here; we don't expect all properties to be set for most use cases
- Main points of interest are how we map between SNS and SQS using the Identifier and a Name value, as we need to get one to refer to the other in cases where we will create both.

This is currently a draft, once we are through feedback, I need to add the JSON schema that matches what we agree too, and then I'll convert to ready


derberg:
@iancooper tbh there are no comments as it is in `Draft` mode. I personally never look at `Draft` PRs as they indicate that it is still in progress not ready for review

jorgebo10:
Hi, any progress done on this. Seems an interesting approach to me.

iancooper:
@derberg 
>an you please remove .DS_Store from PR? also Conventions.md is probably from another PR. Also since some time, bindings require JSON Schemas, please add them

Will do, I had left the schema to get discussion of the proposal, before doing that work. It was originally why I put it into draft. It might be worth thinking about process here.

>as currently these bindings are empty, will you also maintain these definitions long term? do you want to become the maintainer of this repo?

Love to. I have a team at Just Eat Takeaway who I can rope in to help make sure we are on top of PRs etc. It could be a lot for one person (and we get the bus problem).

>I'm really sorry for keeping it waiting, for just such a short comment for 4 months
All good, I owe you picking up on Spec 3.0 and coming to some more meetings. We do what we can.

derberg:
> Will do, I had left the schema to get discussion of the proposal, before doing that work. It was originally why I put it into draft. It might be worth thinking about process here.

ok, that makes sense, for initial review markdown is critical

> Love to. I have a team at Just Eat Takeaway who I can rope in to help make sure we are on top of PRs etc. It could be a lot for one person (and we get the bus problem).

Perfect ❤️ 

---
if you know other folks outside your company and use cases using SNS/SQS that you can reach for giving feedback in this PR, please contact them, also on social media. I will also ping community on AsyncAPI social accounts

sqs/README.md
```diff
@@ -2,36 +2,237 @@
...
+|<a name="channelBindingObjectBindingVersion"></a>`bindingVersion` | string | **Optional**, defaults to `latest`. The version of this binding.|
+
+### Schemas
+
+#### Queue
+|Field Name | Type | Description|
+|---|:---:|---|
+|<a name="queueObjectRef"></a>$ref | `string` | Allows for an external definition of a queue. The referenced structure MUST be in the format of a [Queue](#queue). If there are conflicts between the referenced definition and this Queue's definition, the behavior is *undefined*.|
+| <a name="queueObjectName"></a>`name` | string | **Required.** The name of the queue. When an [SNS Operation Binding Object]() references an SQS queue by name, the identifier should be the one in this field.|
+| <a name="queueObjectType"></a>`type` | string one of: fifo, standard | **Required.**  Is this a FIFO queue or a standard queue? |
```
zekth, line 51: Any reason to use a string field in here instead of a Boolean for `FifoQueue` ? Aws defines it as one: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sqs-queue.html

sqs/README.md
```diff
@@ -2,36 +2,237 @@
...
+|Field Name | Type | Description|
+|---|:---:|---|
+| <a name="redrivePolicyObjectDeadLetterQueue"></a>`deadLetterQueue` |[Identifier](#identifier)| **Required.** The SQS queue to use as a dead letter queue (DLQ) |
+| <a name="redrivePolicyObjectMaxReceiveCount"></a>`maxReceiveCount` |integer| **Optional.** The SQS queue to use as a dead letter queue (DLQ) 
+
+#### Statement
+|Field Name | Type | Description|
+|---|:---:|---|
+| <a name="channelBindingPolicyStatementObjectEffect"></a>`effect` | string |**Required.** Either "Allow" or "Deny"|
+| <a name="channelBindingPolicyStatementObjectPrincipal"></a>`principal` | string or array of string |**Required.** The AWS account or IAM user to whom this statement applies|
```
zekth, line 81: Same remark as SNS.

sns/README.md
```diff
@@ -15,23 +15,288 @@ Current version is `0.1.0`.
...
+
+#### Consumer
+
+| Field Name | Type | Description |
+|---|:---:|---|
+| <a name="operationBindingObjectProtocol"></a>`protocol` |string, one of: http, https, email, email-json, sms, sqs, application, lambda, firehose | **Required.** What protocol will this endpoint receive messages by.|
+| <a name="operationBindingObjectEndpoint"></a>`endpoint` |[identifier](#identifier)| **Required.** Where are messages being delivered to?|
+| <a name="operationBindingObjectFilerPolicy"></a>`filterPolicy` | [filterPolicy](#filter-policy) | **Optional.** Only receive a subset of messages from the channel, determined by this policy.|
+| <a name="operationBindingObjectRawMessageDelivery"></a>`rawMessageDelivery` | boolean | **Required.** If *true* AWS SNS attributes are removed from the body, and for SQS, SNS message attributes are copied to SQS message attributes. If *false* the SNS attributes are included in the body. |
+| <a name="operationBindingObjectRedrivePolicy"></a>`reDrivePolicy` | [redrivePolicy](#redrive-policy) | **Optional.** Prevent poison pill messages by moving un-processable messages to an SQS dead letter queue. |
```
zekth, line 118:
```suggestion
| <a name="operationBindingObjectRedrivePolicy"></a>`redrivePolicy` | [redrivePolicy](#redrive-policy) | **Optional.** Prevent poison pill messages by moving un-processable messages to an SQS dead letter queue. |
```
There is the casing issue over the spec. I think `redrivePolicy` is the good format.

Conventions.md
```diff
@@ -0,0 +1,301 @@
...
+
+The Binding Extensions format works with tools because it indicates where in the specification the binding should be used via a Hook. Using a Hook signals clearly to the user of an extension on which of Channel, Operation or Message the binding can be used. Implementers may choose to use author a binding so that the only fields on the binding are those appropriate to the object identified by the Hook. It also makes it possible for tools to determine correctness of an AsyncAPI file by ensuring that the binding is only used where permitted.
+
+### **Item** 2 Prefer JSON for Bindings
+Prefer the use of JSON to define a Binding. 
+
+[The AsyncAPI specification is in JSON](https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#format), so using JSON to define the Binding assures consistency of the format of the two. 
+
+Whilst both JSON and YAML are human readable, YAML is more readable. As [YAML is a superset of JSON](https://yaml.org/spec/1.2/spec.html#id2759572) users may prefer to write AsyncAPI files in YAML, benefitting from the improved readability.
```
zekth, line 24:
```suggestion
While both JSON and YAML are human readable, YAML is more readable. As [YAML is a superset of JSON](https://yaml.org/spec/1.2/spec.html#id2759572) users may prefer to write AsyncAPI files in YAML, benefitting from the improved readability.  
```

zekth (changes requested):
Couple nits here and there at the first pass.
As previously mentionned the `.DS_Store` file needs to be removed and the json schemas generated.

I also mentionned this PR to @Kong folks as we are inclined to incorporate those specs in our platforms.

Kamran64:
Hey, I see there's not been traction in this PR in a while. When can we expect these bindings (particularly SNS as that's what I need 😁)  to be included in the Async spec/what else needs to be done?

iancooper:
Kamran,

I’ll try to pick this up with my team, we need to add the schema, but really we were waiting for comments from others before we went ahead with that

>

Kamran64:
@iancooper - Thanks for the update, who do you need reviews from? This Async team directly or people in the community who have worked with SNS before? If the latter I can see if I can loop some people in to share their thoughts.

sns/README.md
```diff
@@ -15,23 +15,288 @@ Current version is `0.1.0`.
...
+```
 
 <a name="operation"></a>
 
 ## Operation Binding Object
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
+This object contains information operation binding in SNS.
+
+We represent SNS producers via a **subscribe** Operation Object. In simple cases this may not require configuration, and can be shown as an empty SNS Binding Object i.e. {} if you need to explicitly indicate how a producer publishes to the channel.
```
jeff9finger, line 91: Maybe I am not understanding something here. Why is an SNS producer represented as a subscribe operation? This seems backwards to me. Shouldn't the SNS producer use a publish operation?

sns/README.md
```diff
@@ -15,23 +15,288 @@ Current version is `0.1.0`.
...
+
+We represent SNS consumers via a **publish** Operation Object. These consumers need an SNS Subscription that defines how they consume from SNS i.e. the protocol that they use, and any filters applied.
```
jeff9finger, line 93: Maybe I am not understanding something here. Why is an SNS consumer represented as a publish operation? This seems backwards to me. Shouldn't the SNS consumer use a subscribe operation?

jeff9finger:
+1

sqs/README.md
```diff
@@ -2,36 +2,237 @@
...
+
+#### Policy
+|Field Name | Type | Description|
+|---|:---:|---|
+| <a name="channelBindingPolicyObjectPolicyStatements"></a>`Statements` | [Statement](#statement) | **Required.** An array of Statement objects, each of which controls a permission for this topic |
+
+#### Redrive Policy
+|Field Name | Type | Description|
+|---|:---:|---|
+| <a name="redrivePolicyObjectDeadLetterQueue"></a>`deadLetterQueue` |[Identifier](#identifier)| **Required.** The SQS queue to use as a dead letter queue (DLQ) |
```
loe-lobo, line 74: @iancooper why DeadLetterQueue is required? Don't you think that the Redrive policy should be a decision of the application and the spec shouldn't enforce it?

sqs/README.md
```diff
@@ -2,36 +2,237 @@
...
+- The producer and consumer both have an AsyncAPI specification file, and the consumer receives commands, for example interop between microservices, and the consumer 'owns' the channel  definition and thus has the SQS Binding on its Channel Object. 
+
+An SQS queue can set up a Dead Letter Queue as part of a Redelivery Policy. To support this requirement, the Channel Binding Object allows you to define both a Queue Object to use as the Channel or target in a *publish* Operation and a Dead Letter Queue. You can then refer to the Dead letter Queue in the Redrive Policy using the Identifier Object and setting the *name* field to match the *name* field of your Dead Letter Queue Object. (If you define the DLQ externally, the Identifier also supports an ARN). 
+
+### Fields
+|Field Name | Type | Description|
+|---|:---:|---|
+| <a name="channelBindingObjectQueue"></a>`queue` | [Queue](#queue)| **Required.** A definition of the queue that will be used as the channel. |
```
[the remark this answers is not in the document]
dpwdec: @jeff9finger I don't think the type is an array. It's of type `Queue` which I think is consistent with source bindings living at this level of the specification. 

> Something like: SNS message --> [SQS 0, SQS 1, SQS 2, ..., SQS n]

The `queues` property does exist on the proposed binding but its part of the binding that sits at the `publish` level for facilitating things like an `sns` configured at the channel level and then multiple consumer queues receiving messages. Example from the proposal below as you cannot link to specific MD lines.

```yaml
sqs:
      queues:
      - name: user-signedup-queue
        type: standard
        receiveMessageWaitTime: 4
        policy:
          statements: 
          - effect : Allow
            principal: *
            action: Sqs:SendMessage
          - effect : Allow
            principal: *
            action: Sqs:ReceiveMessage 
```

sqs/README.md
```diff
@@ -2,36 +2,237 @@
...
+            rawMessageDelivery: true  
+            filterPolicy:
+              attributes:
+                reason: 
+                  anything-but: password-reset
+            reDrivePolicy:
+              deadLetterQueue:
+              name: user-signedup-queue-dlq
+        sqs:
+          queues:
```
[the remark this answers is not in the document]
dpwdec: Hope my comment above addressed this. At the `publish` level you can configure multiple queues that receive data, but at the channel level you should only configure a single queue because this is the primary logical pipe through which data flows. For the example a point to point SQS model (in the single channel level queue example) as opposed to SNS -> SQS pub sub model (possible one to many messaging).

rngtng:
How can I help to get this PR forward? I'm about to spec messages for SNS/SQS and would love to make use of this

dpwdec:
@rngtng Apart from the question raised by @loe-lobo about changing DLQ to not be required. I think it's good to go. After that is changed we could move towards merging it to allow standardisation in other tooling to be built.

iancooper:
@dpwdec Merged here

dpwdec:
@rngtng @loe-lobo @jeff9finger @Kamran64 @zekth @derberg I believe I've now addressed everyone's comments and concerns so I'm going to move onto writing the schemas for this PR so that we can facilitate other AsyncAPI tooling that relies on these bindings.

dpwdec:
@fine-project-kata I'm not sure I understand what you are asking? Are you querying why we can have a different SNS/SQS name from the channel in which that piece of infrastructure is configured?

dpwdec:
@fine-project-kata 

So, I think get the gist of what you are asking.

In terms of a case where the resource in question (for this example SQS) varies by environment I would say that this falls *outside* of the scope to be directly documented in AsyncAPI, it can be, but as you've pointed out above it becomes complex and verbose, so this is an implementation detail of one's use case and should be handled as such. To clarify, at JET we have a similar architectural design that you've raised above. For integrators using the AsyncAPI for documentation what they need to understand is that there is a channel that exists that serves said data, the specifics of data might vary from environment to environment, but the shape and semantic meaning/use of the data (or command) will stay consistent across environments. 

So if we were to expand the example above. The canonical channel which serves data is `user-signedup`. This channel is an abstraction that gives out data related to this action, command, projection etc. *you get user signed up data as appropriate when you consumer from it*. The `sqs` binding within that channel is then an implementation specific infrastructure detail and should be named as such with `{env}-user-signedup`. Then, one should use the [AsyncAPI templater tooling](https://www.asyncapi.com/docs/tools/generator/template) provided by the org to parse this into a usable infrastructure definition for whatever deployment mechanism you might use.

In this case, we keep a strong separation of concerns where we can easily replace `sqs` with `kafka` or `mqtt` without effecting the meaning of the data or the channel, our integrators can simply register for the same channel and be aligned on data and uses, futhermore owners then have fewer channels to maintain. Fundamentally: 

> **The channel design should follow architecture not implementation.**

For the very simple use case where there is only a single queue in a single environment the `name` field can appear redundant but for anything more complex, such as documenting legacy infrastructure, or templating out infrastructure across multiple environments or even technologies it becomes essential.

The same comments above apply for DLQ. I would assert that DLQ is most often an internal implementation detail for integrators and should not be separately documented from the main queue because its not a data source you are expecting anyone else to integrate with, it exists to facilitate error handling for the main queue. (If a user is not using DLQ this way but instead as a feature for message filtering or something I would say that we should not be facilitating this as it clearly goes against the infrastructure designs of the tool).

Most importantly I think the structure you are proposing above where you separate all queues and DLQs by explicit channels can be accomplished with above spec, you would need to explicitly state the name of the resource, however, you could always implementing your own parser using the templater to infer name from channel. So if you don't want to follow the structure you are suggesting there's nothing really stopping you given the above binding proposal. But I believe the requirement for name in general as a way to disambiguate channel and infrastructure concerns is correct.

fine-project-kata:
@dpwdec thanks for explanation , it makes sense form this perspective. It will be consistent with all spec .

iancooper:
@dpwdec Should be merged, sorry for the delay

KhudaDad414 (changes requested):
IMO the changes are too much for a single PR.
Great work.

dpwdec:
@KhudaDad414 Thank you for all of the comments and reviews, I will try and address all of these, I appreciate you taking the time to look through it all carefully. 

In terms of the PR being too large will this block merging?

KhudaDad414:
> In terms of the PR being too large will this block merging?

I don't think it will. But since SNS and SQS are being added as two different protocols I guess it makes sense to have them in two separate PRs. plus it would be a lot easier to review.

unless there is a reason behind having both as a single PR :)

sns/README.md
```diff
@@ -15,23 +15,288 @@ Current version is `0.1.0`.
...
+We represent an AsyncAPI Channel with a Topic in SNS. The bindings here allow definition of a topic within SNS. We provide properties on the binding that allow creation of a topic in infrastructure-as-code scenarios. Be aware that although the binding offers that flexibility, it may be more maintainable to specify properties such as SNS Access Control Policy outside of AsyncAPI.
+
+SNS supports many optional properties. To mark a channel as SNS, but use default values for the channel properties, just use an empty object {}. 
+
+### Fields
+
+|Field Name | Type | Description|
+|---|:---:|---|
+| <a name="channelBindingObjectName"></a>`name` | string | **Required.** The name of the topic. Can be different from the channel name to allow flexibility around AWS resource naming limitations.|
+| <a name="channelBindingObjectFifo"></a>`type` | [ordering](#ordering)| **Optional.** By default, we assume an unordered SNS topic. This field allows configuration of a FIFO SNS Topic.  |
```
dpwdec, line 33: Perhaps we should rename this property to `ordering` instead of `type` as type is a reserved keyword in many languages that might want to make use of this schema and it's also a reserved word in JSON schema and can easily become confusing. It's quite overloaded.

VisualBean (commented):
"Attributes" makes me somewhat iffy. 
"Map of string or object."

That's really hard to implement in pretty much all languages, for tooling, Is that how it is today in SNS? Just a map of strings and maps? (Objects being maps)

sns/README.md
```diff
@@ -15,23 +15,288 @@ Current version is `0.1.0`.
...
+| <a name="channelBindingDeliveryPolicyObjectNumMinDelayRetries"></a>`numMinDelayRetries` | integer | **Optional.** The number of immediate retries (with delay) |
+| <a name="channelBindingDeliveryPolicyObjectNumMaxDelayRetries"></a>`numMaxDelayRetries` | integer | **Optional.** The number of post-backoff phase retries, with the maximum delay between retries |
+| <a name="channelBindingDeliveryPolicyObjectBackoffFunction"></a>`backoffFunction` | string, one of: arithmetic, exponential, geometric or linear | **Optional.** The algorithm for backoff between retries |
+| <a name="channelBindingDeliveryPolicyObjectMaxReceivesPerSecond"></a>`maxReceivesPerSecond` | integer | **Optional.** The maximum number of deliveries per second, per subscription |
+
+#### Filter Policy
+|Field Name | Type | Description|
+|---|:---:|---|
+| <a name="filterPolicyObjectAttributes"></a>`attributes` |Map(string, array or string or object) | **Required.** A map of a message attribute to an array of possible matches. The match may be a simple string for an exact match, but it may also be an object that represents a constraint and values for that constraint  |
```
VisualBean, line 138: This makes me a bit iffy. 

A map of pretty much anything. Is really hard to implement tooling around in typed languages. 

Is this how it is today? A map of arrays, objects or strings, from Amazons side?

dpwdec: @VisualBean based on the [example filter policies that Amazon documents](https://docs.aws.amazon.com/sns/latest/dg/example-filter-policies.html), it does seem that way, its not as simple as just a flat mapping either with nested objects, it seems like its some form of arbitrary JSON.

sns/json_schemas/operation.json
```diff
@@ -0,0 +1,203 @@
...
+        "arn": {
+          "type": "string",
+          "description": "The target is an ARN. For example, for SQS, the identifier may be an ARN, which will be of the form: arn:aws:sqs:{region}:{account-id}:{queueName}"
+        },
+        "name": {
+          "type": "string",
+          "description": "The endpoint is identified by a name, which corresponds to an identifying field called 'name' of a binding for that protocol on this publish Operation Object. For example, if the protocol is 'sqs' then the name refers to the name field sqs binding. We don't use $ref because we are referring, not including."
+        }
+      }
+    },
```
dpwdec, line 64: Looking at the AWS documentation for redrive policies it doesn't seem like the target resource to be used for redriving messages is an object but is instead a string that could be an ARN, URL, Name etc.

For example in cloud formation the redrive policy definition looks like this:
```
"{ \"deadLetterTargetArn\": \"arn:aws:sqs:us-east-2:111122224444:my-dlq\", \"maxReceiveCount\": 4 }"
```

Which appears to be [fairly representative](https://awslabs.github.io/serverless-rules/rules/sqs/redrive_policy/) of how a lot of [infrastructure tooling deals with defining redrives](https://awslabs.github.io/serverless-rules/rules/sns/redrive_policy/).

Is there a reason we are diverging from the common standards for defining redrives with a separate identifier object  that has specific properties?

gokerakc: In my opinion, it's better to keep the definitions as simple as possible at the beginning. If most infra tooling only uses **deadLetterTargetArn** and **maxReceiveCount** for the Redrive Policy then there is no harm in starting with that properties.

As a result, we just need a model that can be interpreted into infra and we can add less complexity by not using comprehensive definitions (if it's applicable).

sns/README.md
```diff
@@ -15,23 +15,288 @@ Current version is `0.1.0`.
...
+
+#### Identifier
```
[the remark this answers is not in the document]
dpwdec: @KhudaDad414 I believe only one would generally be used. But I guess there are potentially use cases where you would want multiple, such as defining the `name` and `arn` for use in different contexts that point to the same resource.

sqs/json_schemas/operation.json
```diff
@@ -0,0 +1,214 @@
...
+         "redrivePolicy": {
+            "deadLetterQueue": {
+              "name": "myQueue_error"
+            },
+            "maxReceiveCount": 15
+         },
+         "policy": {
+            "statements": [
+              {
+                "effect": "deny",
```
KhudaDad414, line 197:
```suggestion
                "effect": "Deny",
```

KhudaDad414 (approved):
Just some small typos. LGTM!

dpwdec:
Thanks @KhudaDad414 nice catches on these, I'm addressing them all and will do my own once over before a merge in any case.

One other thing has come up while discussing this: 

How do you feel about the objects _within_ bindings being extensible? I notice that in a lot of other bindings the objects under the root binding don't have a `patternProperties` entry that allows for `x-` extensions. However, I'm not convinced this was intentional and perhaps more of just a piece of functionality that hasn't been considered?

Especially with these `sqs` and `sns` bindings that mirror the AWS APIs somewhat the ability to have codified extension fields on the objects themselves seems important as a way to offer better usability and reduce very verbose specs.

For example, if we have an `sqs` operation binding with multiple queues configured like so:
```yaml
sqs:
  queues:
    - name: "my-queue"
      type: standard
      maxReceiveCount: 2
   - name: "my-queue_error"
      maxReceiveCount: 10
```

And I would like to add an internal extension property to the dead letter queue (which I think would be a common enough use case given different organisations often have internal standards for infrastructure deployment). It would make sense and be more readable for users to do this directly on the `Queue` object in the list. In its current form the PR describes a spec that requires `x-` extension properties to be limited to the root object which leads to a very verbose configuration like this:
```yaml
sqs:
  queues:
    - name: "my-queue"
      type: standard
      maxReceiveCount: 2
   - name: "my-queue_error"
      maxReceiveCount: 10
  x-mySqsOperationBindingExtension:
    queues:
      - some: no-op # don't want add custom configuration to the first queue element so need some no-op value
      - myInternalQueueConfigurationProperty: "some value"
```

As opposed to the much simpler configuration offered by `patternProperties` with `x-` extensions on the underlying objects:
```yaml
sqs:
  queues:
    - name: "my-queue"
      type: standard
      maxReceiveCount: 2
   - name: "my-queue_error"
      maxReceiveCount: 10
      x-myInternalQueueConfigurationProperty: "some value"
```

What do you think? Do you have any strong objections to allowing these sorts of extensions in the schema? It seems sensible to me and would definitely facilitate the real world use cases I've seen of SNS/SQS in AsyncAPI given the structure of the bindings.

KhudaDad414:
Hey @dpwdec,
IMO this is something that the maintainers of SNS and SQS bindings should decide.
I am here to make sure that the documentation is up-to-date with the JSON schemas. Don't know a lot about SNS and SQS. 🤷 Sorry.
BTW, I am not sure who is going to maintain these. 
We should add them to [CODEOWNERS](https://github.com/asyncapi/bindings/blob/master/CODEOWNERS) before merge.

dpwdec:
@KhudaDad414 I mean, I will probably act as the maintainer as I've been driving forward the specification for it, so I'll add myself. I guess I'm also acting as the primary user as we are using these draft bindings already internally substantially so I think I have a reasonable idea of what makes sense to include.

dpwdec:
@KhudaDad414 I think all those changes are merged in now. If you're all happy feel free to merge. 🫡

sqs/README.md
```diff
@@ -2,36 +2,237 @@
...
+| <a name="channelBindingObjectQueue"></a>`queue` | [Queue](#queue)| **Required.** A definition of the queue that will be used as the channel. |
+| <a name="channelBindingObjectDLQ"></a>`deadLetterQueue` | [Queue](#queue)| **Optional.** A definition of the queue that will be used for un-processable messages. |
+|<a name="channelBindingObjectBindingVersion"></a>`bindingVersion` | string | **Optional**, defaults to `latest`. The version of this binding.|
+
+### Schemas
+
+#### Queue
+|Field Name | Type | Description|
+|---|:---:|---|
+|<a name="queueObjectRef"></a>$ref | `string` | Allows for an external definition of a queue. The referenced structure MUST be in the format of a [Queue](#queue). If there are conflicts between the referenced definition and this Queue's definition, the behavior is *undefined*.|
```
VisualBean, line 49: Defined where? Is it a link, or is 'external' meaning it should go into `Components`? which does not define this behaviour

gokerakc: I think we can remove the **queueObjectRef** because that looks like a **URI Reference** explained in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-4.1). This can be either **relative** or **absolute** reference.

URI References can be used for any object when it's needed. Any software libraries which serialise/deserialise JSON or YAML files must provide the option (or the default) to handle the references.

Because the **URI Reference** property (**$ref**) is a standard and can be used in any object I think it's unnecessary to specify the **$ref** property again in the **Queue** object.

derberg:
@dpwdec thanks for all the hard work!!! I see no objection from other participants in this PR to make you a maintainer of these bindings

Just please remove `.DS_Store` from PR

---

There is one concern I have, the conventions file, it is not related to these new bindings only but sets a conventions on all the bindings, which should be done in a separate PR imho, to not mix contexts. Conventions are something that need to be respected, thus part of contribution guide and review process. So to make sure that happens, separate discussion and 👍🏼 from other bindings owners is a must imho

dpwdec:
Hi @derberg! Thanks for taking the time to look through it all :D I've removed `.DS_Store` now and migrated `conventions.md` to a separate PR which I will expand for further discussion. So I think we may be good to go?

JaredAAT:
Is this likely to get merged soon?

dpwdec:
@JaredAAT As far as I am aware everything is complete. It's just waiting to be merged.

derberg:
@dpwdec sorry for another question but I just want to be 100% sure before we merge

- @iancooper is the creator of the PR but at some point of time you took over. I tried to follow but could not find a comment when you folks agree on that
- I'm also asking because at the moment I see that you will be the only maintainer for the binding. We do allow it but it usually happens when we know the contributor form discussions, especially [Slack](https://asyncapi.com/slack-invite) presence and we know that in case of emergency there are some ways to contact in a way different than GitHub, and they can also easily reach us in case help is needed. I hope I explain myself well. So in case of recent addition of `jms` binding a totally new community member became maintainer because another existing maintainer agreed to join with him

Basically trying to figure what will happen the next day we merge this 😅

iancooper:
> @dpwdec sorry for another question but I just want to be 100% sure before we merge
> 
>     * @iancooper is the creator of the PR but at some point of time you took over. I tried to follow but could not find a comment when you folks agree on that

Hi @derberg, thanks for asking. I work with @dpwdec. He is part of a team at Just Eat Takeaway I provide advice to, who are implementing AsyncAPI for us. I ran out of time to pursue this PR, so I delegated it to the team instead. I am happy to keep being informed and help out, but the team has more time for the detail here than I do right now. 

>     * I'm also asking because at the moment I see that you will be the only maintainer for the binding. We do allow it but it usually happens when we know the contributor form discussions, especially [Slack](https://asyncapi.com/slack-invite) presence and we know that in case of emergency there are some ways to contact in a way different than GitHub, and they can also easily reach us in case help is needed. 

Happy to keep being a maintainer with @dpwdec to avoid a single-point of failure. We might be able to find some other folks in the team too.

dpwdec:
@derberg Seconding what's said above, more than happy to provide support for SNS/SQS bindings in tandem with @iancooper or any others who feel they have the expertise or time to provide the support as well. I'm also on the slack now and can provide more contact details to other AsyncAPI maintainers if required. 👍

derberg:
@dpwdec sorry if I sounded like you are not welcome or something like that. It’s just that the problem with bindings we have that they are basically “micro specifications for AsyncAPI” so they do not change often, very rare, so there is no proper path for potential maintainer to feel the AsyncAPI community spirit before we invite to the org. 

So basically since last few months where AnypointMQ spec maintainers had to change we were afraid we just invite someone who we will not be really able to contact with in case of problems, someone that is not known. I hope you know what I mean. So in JMS we luckily had existing codeowner volunteer to help the other “fresh” maintainer. 

Governance of AsyncAPI Initiative works in the way that any maintainer can become a [TSC member](https://www.asyncapi.com/community/tsc) without any secret invitation to secret club 😄 so good to know that the person we appoint for maintainer role is easy to reach on different channels in case of emergency.

anyway, super happy to see @iancooper already commenting on this. He is a great guy that was very active some time ago so if you folks work together and are both fine to be co-maintainers that is a perfect situation 🍺 

@dpwdec please update the CODEOWNERS and add Ian

After we merge this PR, feel free to add your names to https://github.com/asyncapi/community/blob/master/MAINTAINERS.yaml through a PR. Do it if you wish to be more involved in the community and help it to grow.

Also, you might be interested in applying for one of our AsyncAPI events this year. We have London CFP opened https://conference.asyncapi.com/ and I bet SQS/SNS talk would be super interesting

iancooper:
All good @derberg I was just conscious of becoming a bottleneck, so I involved the team here, and @dpwdec has done a ton of heavy lifting. He is using my fork, so I see the PRs before merging.

dpwdec:
@derberg No problem at all, I didn't feel it was unwelcoming! As you've said above there are can and will be issues and you need to make sure there are some contingencies in place for when that happens. I'll update CODEOWNERS tomorrow morning and take a look at the conference as well! 🙌

dpwdec:
@derberg 'CODEOWNERS` has been updated now 😊

derberg (approved):
LGTM, thanks for all the hard work

@iancooper @dpwdec 

so this binding will work with AsyncAPI prior 3.0

once this is merged, please look into https://github.com/asyncapi/bindings/issues/182. We need you to inspect if there will be any changes needed to make binding work with `3.0`

derberg:
@KhudaDad414 you need to approve again as JSON Schemas guardian 😄

KhudaDad414 (approved):
LGTM!

KhudaDad414:
Great effort @iancooper @dpwdec 🙇

PR #85, opened 2021-09-16
feat: add solace binding

### LINKED ISSUE #80: Feature request: Solace bindings
#### Reason/Context
Please try answering few of those questions
- Why we need this improvement?
  At Solace we are expanding our use of AsyncAPI from code generation to discovery and beyond. We need a way to represent things like queue configuration.
- How will this change help?
 It will allow us to represent the details needed to describe more technical features of messaging with Solace.
- What is the motivation?
We see that there already exist bindings for protocols/products such as IBMMQ and Kafka, and we would like to do something similar for Solace. It will allow us to use AsyncAPI in more use cases.

#### Description
Please try answering few of those questions
- What changes have to be introduced?
 We will provide a new page for the Solace bindings.
- Will this be a breaking change?
 No.
- How could it be implemented/designed?
Here is the proposed page: https://github.com/damaru-inc/bindings/tree/solace-bindings-initial/solace

damaru-inc:
**Description**

This describes bindings for the Solace Messaging Format protocol.

**Related issue(s)**
Resolves #80


solace/README.md
```diff
@@ -0,0 +1,121 @@
...
+          - destinationType: queue
+            queue:
+              name: CreatedHREvents
+              topicSubscriptions:
+              - person/*/created
+          - destinationType: queue
+            queue:
+              name: UpdatedHREvents
+              topicSubscriptions:
+              - person/*/updated
```
magicmatatjahu, line 102: Probably there is a little bug :)

```suggestion
          bindingVersion: 0.2.0
          bindings:
            - destinationType: queue
              queue:
                name: CreatedHREvents
                topicSubscriptions:
                - person/*/created
            - destinationType: queue
              queue:
                name: UpdatedHREvents
                topicSubscriptions:
                - person/*/updated
```

solace/operation.json
```diff
@@ -0,0 +1,81 @@
...
+                "items": {
+                  "type": "string"
+                }
+              },
+              "exclusive": {
+                "type": "boolean"
+              }
+            }
+          }
+        }
```
magicmatatjahu, line 49: If `queue` is only valid with `destinationType: queue` then we can use the `oneOf` to have more restricted validation:

```json
      "items": {
        "type": "object",
        "properties": {
          "deliveryMode": {
            "type": "string",
            "enum": [
              "direct",
              "persistent"
            ]
          }
        },
        "oneOf": [
          {
            "properties": {
              "destinationType": {
                "type": "string",
                "const": "queue",
                "description": "If the type is queue, then the subscriber can bind to the queue, which in turn will subscribe to the topic as represented by the channel name."
              },
              "queue": {
                "type": "object",
                "properties": {
                  "name": {
                    "type": "string",
                    "description": "The name of the queue"
                  },
                  "topicSubscriptions": {
                    "type": "array",
                    "description": "The list of topics that the queue subscribes to.",
                    "items": {
                      "type": "string"
                    }
                  },
                  "exclusive": {
                    "type": "boolean"
                  }
                }
              }
            }
          },
          {
            "properties": {
              "destinationType": {
                "type": "string",
                "const": "topic",
                "description": "If the type is queue, then the subscriber can bind to the queue, which in turn will subscribe to the topic as represented by the channel name."
              }
            }
          }
        ]
      }
```

magicmatatjahu (commented):
Some suggestions :)

solace/operation.json
damaru-inc: That's a good idea, I'll consider it.

Mrc0113:
Thanks for putting this together @damaru-inc. I think it looks good and looks extendible going forward should we need to add other solace specific things based on feedback from users.

derberg (dismissed):

derberg:
@fmvilas ping 😉 
@damaru-inc hey, can you update your branch with the latest master and add this new binding to just newly created https://github.com/asyncapi/bindings/blob/master/CODEOWNERS (so yeah, it is all done with a single PR)

damaru-inc:
> @fmvilas ping 😉 @damaru-inc hey, can you update your branch with the latest master and add this new binding to just newly created https://github.com/asyncapi/bindings/blob/master/CODEOWNERS (so yeah, it is all done with a single PR)

Done!

derberg (approved):

derberg:
@damaru-inc I see you added Cameron. He did not contribute yet to bindings repo, but as he is already maintainer for other projects and TSC member I guess it is fine. You are ok that he will be able to approve changes in the binding?

derberg:
@damaru-inc I don't think it is a big problem as long as @CameronRushton is aware it is fine. He already is in the TSC and we are talking here about only `solace` binding, so I think it is right

@fmvilas anything from your side on the subject?

CameronRushton:
> I don't think it is a big problem as long as @CameronRushton is aware it is fine.

I'm good with it

fmvilas:
All good from my side 👍

fmvilas (approved):

PR #107, opened 2022-01-12
feat: add topicSubscriptions to direct destinations.

damaru-inc:
**Description**
The solace binding should have a topicSubscriptions field in the topic destination, just like we have for queues.

Also something was fixed in the schema. The schema had a property called 'exclusive' which should have been 'accessType.'


KhudaDad414 (dismissed):
JSON schemas and Descriptions seem to comply. 🚀

CameronRushton (approved):
LGTM

PR #115, opened 2022-02-04
feat: add schema registry infos to Kafka binding

lbroudoux:
**Description**

As of today, there's no way using `0.2.0` version of Kafka bindings to provide information on whether a Schema Registry was/has to be used to provide/gather schema information for messages published on topics.

However, such information is required to help on two use-cases: 
1. Inform API providers to enable schema registry connection and pick the right vendor serdes,
2. Inform API consumers on the same things but also how to bypass if schema registry is unavailable

Use-cases and needed informations have been discussed with @nictownsend and @dalelane. This PR is aggregating validated propositions. The proposition is to add following attributes:

At the `server` level:
- `schemaRegistryUrl` - API URL for the Schema Registry used when producing Kafka messages
- `schemaRegistryVendor` - he vendor of Schema Registry and Kafka serdes library that should be used

At the `operation` level:
- `schemaIdLocation` - Tells where the id of schema is stored (e.g. `header` or `payload`)
- `schemaIdPayloadEncoding` -  Number of bytes or vendor specific values when schema id is encoded in payload
- `schemaLookupStrategy` - Freeform string for any naming strategy class to use. Clients should default to the vendor default if not supplied.

**Related issue(s)**

This PR is related to #40 and #41. It has been initially discussed into PR #55 but finally close due to inactivity.

### DIFF
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -40,14 +40,39 @@ servers:
 
 ## Channel Binding Object
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
+This object contains information about the channel representation in Kafka (eg. a Kafka topic).
 
+##### Fixed Fields
+
+Field Name | Type | Description | Applicability [default] | Constraints
+---|:---:|:---:|:---:|---
+<a name="channelBindingObjectTopic"></a>`topic` | string | Kafka topic name if different from channel name. | OPTIONAL | -
+<a name="channelBindingObjectPartitions"></a>`partitions` | integer | Number of partitions configured on this topic (useful to know how many parallel consumers you may run). | OPTIONAL | Must be positive
+<a name="channelBindingObjectReplicas"></a>`replicas` | integer | Number of replicas configured on this topic. | OPTIONAL | MUST be positive
+<a name="channelBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed. | OPTIONAL [`latest`] | -
+
+This object MUST contain only the properties defined above.
+
+##### Example
+
+This example is valid for any Confluent compatible schema registry. Here we describe the implementation using the first 4 bytes in payload to store schema identifier.
+
+```yaml
+channels:
+  user-signedup:
+    bindings:
+      kafka:
+        topic: 'my-specific-topic-name'
+        partitions: 20
+        replicas: 3
+        bindingVersion: '0.3.0'
+```
 
 <a name="operation"></a>
 
 ## Operation Binding Object
 
-This object contains information about the operation representation in Kafka.
+This object contains information about the operation representation in Kafka (eg. the way to consume messages)
 
 ##### Fixed Fields
 
@@ -55,9 +80,6 @@ Field Name | Type | Description | Applicability [default] | Constraints
 ---|:---:|:---:|:---:|---
 <a name="operationBindingObjectGroupId"></a>`groupId` | [Schema Object][schemaObject] | Id of the consumer group. | OPTIONAL | -
 <a name="operationBindingObjectClientId"></a>`clientId` | [Schema Object][schemaObject] | Id of the consumer inside a consumer group. | OPTIONAL | -
-<a name="operationBindingObjectSchemaIdLocation"></a>`schemaIdLocation` | string | If a Schema Registry is used when performing this operation, tells where the id of schema is stored (e.g. `header` or `payload`). | OPTIONAL | MUST NOT be specified if `schemaRegistryUrl` is not specified at the Server level
-<a name="operationBindingObjectSchemaIdPayloadEncoding"></a>`schemaIdPayloadEncoding` | string | Number of bytes or vendor specific values when schema id is encoded in payload (e.g `confluent`/ `apicurio-legacy` / `apicurio-new`). | OPTIONAL | MUST NOT be specified if `schemaRegistryUrl` is not specified at the Server level
-<a name="operationBindingObjectSchemaLookupStrategy"></a>`schemaLookupStrategy` | string | Freeform string for any naming strategy class to use. Clients should default to the vendor default if not supplied. | OPTIONAL | MUST NOT be specified if `schemaRegistryUrl` is not specified at the Server level
 <a name="operationBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed. | OPTIONAL [`latest`] | -
 
 This object MUST contain only the properties defined above.
@@ -67,7 +89,7 @@ This object MUST contain only the properties defined above.
 ```yaml
 channels:
   user-signedup:
-    publish:
+    subscribe:
       bindings:
         kafka:
           groupId:
@@ -76,9 +98,6 @@ channels:
           clientId:
             type: string
             enum: ['myClientId']
-          schemaIdLocation: 'payload'
-          schemaIdPayloadEncoding: 'apicurio-new'
-          schemaLookupStrategy: 'TopicIdStrategy'
           bindingVersion: '0.3.0'
 ```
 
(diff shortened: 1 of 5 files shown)


dalelane (commented):
Someone using schemas (and a schema registry) could use that to serialize a message key, the message payload, or both. 

Is the assumption here that schema-related bindings settings relate to message payload only?

kafka/README.md
```diff
@@ -33,11 +51,14 @@ This object contains information about the operation representation in Kafka.
...
-<a name="operationBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
+Field Name | Type | Description | Applicability [default] | Constraints
+---|:---:|:---:|:---:|---
+<a name="operationBindingObjectGroupId"></a>`groupId` | [Schema Object][schemaObject] | Id of the consumer group. | OPTIONAL | -
+<a name="operationBindingObjectClientId"></a>`clientId` | [Schema Object][schemaObject] | Id of the consumer inside a consumer group. | OPTIONAL | -
+<a name="operationBindingObjectSchemaIdLocation"></a>`schemaIdLocation` | string | If a Schema Registry is used when performing this operation, tells where the id of schema is stored (e.g. `header` or `payload`). | OPTIONAL | MUST NOT be specified if `schemaRegistryUrl` is not specified at the Server level
+<a name="operationBindingObjectSchemaIdPayloadEncoding"></a>`schemaIdPayloadEncoding` | string | Number of bytes or vendor specific values when schema id is encoded in payload (e.g `confluent`/ `apicurio-legacy` / `apicurio-new`). | OPTIONAL | MUST NOT be specified if `schemaRegistryUrl` is not specified at the Server level
```
dalelane, line 59: My personal preference is towards describing the implementation (e.g. first four bytes, or first eight bytes) rather than use named vendors. I think that would be easier to maintain, as vendors are prone to change what they do or add new options. 

Plus this way we won't end up with multiple ways of describing the same thing (i.e. where multiple vendors have a similar approach)

kafka/README.md
```diff
@@ -33,11 +51,14 @@ This object contains information about the operation representation in Kafka.
...
-Field Name | Type | Description
----|:---:|---
-<a name="operationBindingObjectGroupId"></a>`groupId` | [Schema Object][schemaObject] | Id of the consumer group.
-<a name="operationBindingObjectClientId"></a>`clientId` | [Schema Object][schemaObject] | Id of the consumer inside a consumer group.
-<a name="operationBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
+Field Name | Type | Description | Applicability [default] | Constraints
+---|:---:|:---:|:---:|---
+<a name="operationBindingObjectGroupId"></a>`groupId` | [Schema Object][schemaObject] | Id of the consumer group. | OPTIONAL | -
+<a name="operationBindingObjectClientId"></a>`clientId` | [Schema Object][schemaObject] | Id of the consumer inside a consumer group. | OPTIONAL | -
+<a name="operationBindingObjectSchemaIdLocation"></a>`schemaIdLocation` | string | If a Schema Registry is used when performing this operation, tells where the id of schema is stored (e.g. `header` or `payload`). | OPTIONAL | MUST NOT be specified if `schemaRegistryUrl` is not specified at the Server level
```
dalelane, line 58: If someone chooses `header` we might want a place to identify what the header is.

kafka/README.md
```diff
@@ -85,7 +109,7 @@ channels:
             key:
               type: string
               enum: ['myKey']
-            bindingVersion: '0.1.0'
+            bindingVersion: '0.3.0'
 ```
 
-[schemaObject]: https://www.asyncapi.com/docs/specifications/2.0.0/#schemaObject
+[schemaObject]: https://www.asyncapi.com/docs/specifications/2.1.0/#schemaObject
```
dalelane, line 115:
```suggestion
[schemaObject]: https://www.asyncapi.com/docs/specifications/2.3.0/#schemaObject
```

derberg:
@lbroudoux do you plan to continue with this one?

kafka/README.md
```diff
@@ -55,7 +76,10 @@ channels:
           clientId:
             type: string
             enum: ['myClientId']
-          bindingVersion: '0.1.0'
+          schemaIdLocation: 'payload'
+          schemaIdPayloadEncoding: 'apicurio-new'
+          schemaLookupStrategy: 'TopicIdStrategy'
```
dratasich, line 81: @lbroudoux thanks for this PR! I'm looking into some bindings for kafka topics too.

regarding schemaLookupStrategy: if its a class, then its not a freeform string, but an enum? is this [Redhat specific](https://access.redhat.com/documentation/en-us/red_hat_integration/2020-q4/html/getting_started_with_service_registry/using-kafka-client-serdes#registry-serdes-concepts-strategy-registry) (couldn't find TopicIdStrategy anywhere else)?

why not specifying the schema name and version for key and value?

lbroudoux: Yes, you're right : `TopicIdStrategy` is one of the strategies used by Apicurio Registry. While this information is useful, I thought it would be cumbersome to manage everything as enum value or FQDN classes...

About using schema name and version, I don't actually know if it's a current option among Ser/Deser libraries ... Maybe it could be another approach but it could be nice specifying a range or inclusion/exclusion pattern for versions in order to support bug fixes.

kafka/README.md
lbroudoux: Hi @dalelane
Is this header value configurable for some Ser/Deser libraries? I thought it was just just dependant on the `schemaRegistryVendor` and couldn't be configured...
If not configurable, I would stick with the header being deduced from the registry vendrod.
What do you think?

kafka/README.md
lbroudoux: That's a good point. We could put the emphasis on the fact that putting the number of bytes is the preferred way of doing things. However, if people are not aware of internal implementations, they could fall back to just specifying the vendor encoding. 

Do you think we should introduce specific notations like `4-bytes` or `8-bytes` or let people put integers ?

lbroudoux:
> @lbroudoux do you plan to continue with this one?

Yep. Sorry for the long time no see. Will have more free cycles now 😉

kafka/README.md
lbroudoux: Resolved with latest commit.

lbroudoux:
Push a new commit to update to latest spec version.

kafka/README.md
lbroudoux, line 81: Thinking about it again, I wonder if those properties are really tied to operation and shouldn't be bound to channel instead. 🤔  I've added them to operation because there already was binding information at that level and adding a new binding at the channel level may certainly make things complex... However, I can't think of a situation where those could be different depending on the operation.
@dalelane Would you have an opinion on that?

lbroudoux: On the same idea, I think it would be nice to add another PR to allow the specification of the topic name as well as the number of configured partitions. IMHO, theses information would strengthen the need for channel-level bindings. Thus we could move there the `schema*` properties.

Any thoughts or opinions?

kafka/README.md
dalelane: I don't have a really strong opinion... I can see arguments for both approaches. I think I lean slightly towards the descriptive enum approach, because it feels slightly more future-proof

kafka/README.md
dalelane: Most schema registries support the Confluent API implementation (in addition to their own APIs), so it's fairly common for apps to use Confluent client libraries pointing at a non-Confluent schema registry, essentially treating it as running in a Confluent-compatibility-mode. The Apicurio schema registry even supports the IBM API and the Confluent API, as well as it's own. ((ref)[https://www.apicur.io/registry/docs/apicurio-registry/2.2.x/getting-started/assembly-intro-to-the-registry.html#_apicurio_registry_capabilities]) 

So I was thinking about how we document this mismatch - where you can't assume you know what a client application is doing just because you know the schema registry vendor. 

I'm actively nit-picking and looking for edge cases to make sure we have something that will cope with all scenarios... but I think your probably right to keep it simple with fewer options. For those scenarios, we can just advise devs to describe their schema registry vendor as how it is being used (e.g. if you're using an Apicurio schema registry with Confluent clients, taking advantage of the Confluent-compatibility APIs, then just describe your schema registry as being Confluent for the purposes of an AsyncAPI document)

kafka/README.md
dalelane: > I wonder if those properties are really tied to operation and shouldn't be bound to channel instead

yeah, I agree - I think that makes more sense

dalelane: > I think it would be nice to add another PR to allow the specification of the topic name as well as the number of configured partitions

There was some discussion of that in https://github.com/asyncapi/spec/issues/623 - I think it'd be really useful, as it's a good pointer to tell you how many clients you may want to run.

lbroudoux: Thinking about it once again ... I think it is even more correct to move schema encoding-related info to the message level. Typically - end even though I don't personally consider this as a best practice - you could have different kinds of Avro records on the same topic. Apicurio also proposes a per-record strategy so I think moving this to the message level will increase the capabilities of describing all kinds of configurations.

lbroudoux: Just pushed a new commit with that change + new channel level information for:
* `topic` in case it's different from the channel name,
* `partitions` that represents the number of configured partitions,
* `replicas` that represents the number of configured replicas for this topic.

dalelane (dismissed):
👍

kafka/README.md
lbroudoux: Keep it simple for the moment and propose to resolve it. Please re-open if any new insights

kafka/README.md
lbroudoux: Propose to resolve it for the moment and keeping it simple.

kafka/json_schemas/channel.json
```diff
@@ -0,0 +1,46 @@
...
+      "^x-[\\w\\d\\.\\-\\_]+$": {
+        "$ref": "https://raw.githubusercontent.com/asyncapi/spec-json-schemas/v2.14.0/schemas/2.4.0.json#/definitions/specificationExtension"
+      }
+    },
+    "properties": {
+      "topic": {
+        "type": "string",
+        "description": "Kafka topic name if different from channel name."
+      },
+      "partition": {
```
KhudaDad414, line 18:
```suggestion
      "partitions": {
```

KhudaDad414 (changes requested):
LGTM! just a small typo.

KhudaDad414 (approved):

dalelane (approved):
nice catch on the typo :)

char0n:
@derberg this PR is made against `master` directly. Can we merge this PR?

fmvilas:
👏 🎉

PR #131, opened 2022-04-26
feat: add support for multiple methods in http bindings

### LINKED ISSUE #2: [FEATURE REQUEST] HTTP with multiple methods
**Is your feature request related to a problem? Please describe.**
 here in asyncapi 2.0.0 we can not add multiple operations like GET-POST-DELETE-PUT using single operation Binding in a single channel.

**Describe the solution you'd like**
 please add multiple operations - (GET, POST, PUT, DELETE) in same channel 
 using Operation Binding like "http"    in asyncapi version 2.0.0
 

**Describe alternatives you've considered**
How it possible to add multiple operations - (GET, POST, PUT, DELETE) in same channel 
 using Operation Binding like "http"    in asyncapi version 2.0.0

**Additional context**
**this is possible**
```smartylighting/streetlights/1/0/action/{streetlightId}/turn/off:
    parameters:
      streetlightId:
        $ref: '#/components/parameters/streetlightId'
    bindings:  
      ws:
        query:
          type: object
          required:
            - companyId
          properties:
            companyId:
              type: number
              minimum: 1
              description: The Id of the company.
          additionalProperties: false
        bindingVersion: '0.1.0'
    publish:
      operationId: turnOff
      bindings:
        http:
          POST:
            type: request
            method: POST
            query:
              type: object
              required:
                - companyname
              properties:
                companyId:
                  type: number
                  minimum: 1
                  description: The Id of the company.
              additionalProperties: false
            bindingVersion: '0.1.0'
          PUT:
            type: request
            method: PUT
            query:
              type: object
              required:
                - companyname
              properties:
                companyId:
                  type: number
                  minimum: 1
                  description: The Id of the company.
              additionalProperties: false
            bindingVersion: '0.1.0'  
      traits:
        - $ref: '#/components/operationTraits/kafka'
      message:
        $ref: '#/components/messages/turnOnOff'```

### PR DESCRIPTION
**Pull request by Souvikns**
**Description**
To support multiple methods in HTTP bindings I am trying to reuse [openAPI's path item object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#path-item-object). Since [openapi](https://www.openapis.org/) is the most widely adopted specification for HTTP, people coming from openAPI would feel right at home. 

Firstly I will try to showcase how the HTTP binding looks with [openapi path item object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#path-item-object) and then try to talk about the complications we might face. 

#### Example 
```yaml
channels:
  /employees:
    bindings:
      http:
        post:
          description: 'create a new Employee data'
          requestBody:
            description: 'Employee to create'
            content:
              application/json:
                schema:
                  type: object
                  properties:
                    name: string
          response:
            '200':
              content:
                application/json:
                  schema:
                    type: object
                    properties:
                      ID: string
            default:
              description: Unexpected Error
              content:
                application/json:
                  schema:
                    type: string
        get:
          description: 'Get employees data as per the id passed'
          parameters:
            - name: ID
              in: query
              reqruied: true
          response:
            '200':
              description: A employee data as per ID
              content:
                application/json:
                  schema:
                    type: object
                    properties:
                      name: string
            default:
              description: Unexpected Error
              content:
                application/json:
                  schema:
                    type: string
```
If you have worked with openAPI before it would look very familiar. With this, we can now define multiple methods in HTTP bindings. Also, any tooling that supports openapi should support this as well. 

#### Some Complications 
Now I don't know if these effects or not, these are just things I think make this a bit complicated to understand 
- **payload** vs **schema** - In AsyncAPI we are using [payload](https://github.com/asyncapi/spec/blob/v2.4.0/spec/asyncapi.md#message-object) to define the message and openapi is using [schema](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#schemaObject) so it could be confusing to people well versed in AsyncAPI specification. 
- [path item object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#path-item-object) in openapi depends on their [server](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#serverObject) object which is a clash with asyncapi [server object](https://github.com/asyncapi/spec/blob/v2.4.0/spec/asyncapi.md#serversObject)

Tagging @fmvilas, @KhudaDad414, and @derberg 

**Related issue(s)**
Fixes #2

### DIFF
--- a/http/README.md
+++ b/http/README.md
@@ -21,8 +21,37 @@ This object MUST NOT contain any properties. Its name is reserved for future use
 <a name="channel"></a>
 
 ## Channel Binding Object
+Describes the http binding operation for a particular channel. This http binding object supports [OpenAPI's Path Item Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#path-item-object) 
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
+##### Example 
+```yaml
+channels:
+  /employees:
+    bindings:
+      http:
+        get:
+          description: 'Get employees data as per the id passed'
+          parameters:
+            - name: ID
+              in: query
+              reqruied: true
+          response:
+            '200':
+              description: A employee data as per ID
+              content:
+                application/json:
+                  schema:
+                    type: object
+                    properties:
+                      name: string
+            default:
+              description: Unexpected Error
+              content:
+                application/json:
+                  schema:
+                    type: string
+```
+
 
 
 <a name="operation"></a>
--- a/http/json_schemas/operation.json
+++ b/http/json_schemas/operation.json
@@ -8,26 +8,36 @@
   "patternProperties": {
     "^x-[\\w\\d\\.\\-\\_]+$": {
       "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/specificationExtension"
-    },
-    "^(GET|POST|PUT|PATCH|DELETE|OPTIONS|HEAD|TRACE)": {
-      "type": "object",
-      "properties": {
-        "query": {
-          "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/schema",
-          "description": "A Schema object containing the definitions for each query parameter. This schema MUST be of type 'object' and have a properties key."
-        },
-        "requestBody": {
-          "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/schema",
-          "description": "A Schema object containing definitions for each request body parameters."
-        },
-        "response": {
-          "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/schema",
-          "description": "A Schema object containing the definitions for each response parameter."
-        }
-      }
     }
   },
   "properties": {
+    "type": {
+      "type": "string",
+      "enum": [
+        "request",
+        "response"
+      ],
+      "description": "Required. Type of operation. Its value MUST be either 'request' or 'response'."
+    },
+    "method": {
+      "type": "string",
+      "enum": [
+        "GET",
+        "PUT",
+        "POST",
+        "PATCH",
+        "DELETE",
+        "HEAD",
+        "OPTIONS",
+        "CONNECT",
+        "TRACE"
+      ],
+      "description": "When 'type' is 'request', this is the HTTP method, otherwise it MUST be ignored. Its value MUST be one of 'GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS', 'CONNECT', and 'TRACE'."
+    },
+    "query": {
+      "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/schema",
+      "description": "A Schema object containing the definitions for each query parameter. This schema MUST be of type 'object' and have a properties key."
+    },
     "bindingVersion": {
       "type": "string",
       "enum": [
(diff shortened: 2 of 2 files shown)


fmvilas:
Great progress. Let's have an eye at what's discussed at https://github.com/asyncapi/spec/issues/94. We have a meeting soon: https://github.com/asyncapi/community/issues/352.

Souvikns:
@fmvilas I updated the PR, let me know your thoughts 😃

fmvilas:
@smoya have a look at this meeting when you have time: https://github.com/asyncapi/community/issues/377. It seems this is not compatible with the work on 3.0.0 as it is right now.

fmvilas:
I have serious doubts we should move this PR forward. In v3, you can have multiple operations of the same kind (send/receive) so there's no need to support multiple HTTP methods in the binding. However, in v2.x, it would be needed. I think this can be left out on purpose to encourage people to migrate to v3.

Or! We make this binding allow both single method and multiple methods depending on the version of the parent document. We make it clear that for v2.x multiple methods are allowed and for v3 only one is allowed. I honestly don't like this solution and would rather favor my proposal above of not supporting multiple methods at all.

Souvikns:
> Or! We make this binding allow both single method and multiple methods depending on the version of the parent document. We make it clear that for v2.x multiple methods are allowed and for v3 only one is allowed. I honestly don't like this solution and would rather favor my proposal above of not supporting multiple methods at all.

We can support multiple methods in bindings for both v2 and v3 since bindings is not actually a part of the spec? But then in v3 we will have two ways of doing the same thing. 

I think the best is to not move forward with this and encourage users to update to v3.

fmvilas:
@derberg @dalelane @char0n @smoya please have a look at the discussion.

smoya:
> In v3, you can have multiple operations of the same kind (send/receive) so there's no need to support multiple HTTP methods in the binding.

> Or! We make this binding allow both single method and multiple methods depending on the version of the parent document. We make it clear that for v2.x multiple methods are allowed and for v3 only one is allowed. I honestly don't like this solution and would rather favor my proposal above of not supporting multiple methods at all.

I agree @fmvilas on both statements.

> I think the best is to not move forward with this and encourage users to update to v3.

I agree @Souvikns on this one.

char0n:
> > In v3, you can have multiple operations of the same kind (send/receive) so there's no need to support multiple HTTP methods in the binding.
> 
> > Or! We make this binding allow both single method and multiple methods depending on the version of the parent document. We make it clear that for v2.x multiple methods are allowed and for v3 only one is allowed. I honestly don't like this solution and would rather favor my proposal above of not supporting multiple methods at all.
> 
> I agree @fmvilas on both statements.
> 
> > I think the best is to not move forward with this and encourage users to update to v3.
> 
> I agree @Souvikns on this one.

The same line of thinking on my end. 

> In v3, you can have multiple operations of the same kind (send/receive) so there's no need to support multiple HTTP methods in the binding. However, in v2.x, it would be needed. I think this can be left out on purpose to encourage people to migrate to v3.

One **wild** thought: bindings can be versioned. So we can theoretically introduce new versions of bindings supporting multiple method, which mostly authors of  AsyncAPI 2.x.y definitions will use, and introduce another version of bindings for  AsyncAPI 3.0 (when it's out).

fmvilas:
> One wild thought: bindings can be versioned. So we can theoretically introduce new versions of bindings supporting multiple method, which mostly authors of AsyncAPI 2.x.y definitions will use, and introduce another version of bindings for AsyncAPI 3.0 (when it's out).

Yeah, the only problem I see with this is maintenance. At some point, someone using v2 will want to add something to the binding but it's not there anymore because it's been removed in the next version. So far we have been fine without it so I wouldn't worry too much about v2 users wanting to have full HTTP support.

derberg:
- work on 3.0 is pretty active
- I see @Souvikns do not have a problem with holding
- I do not see a great push from community to have this. Some people asked if REST APIs can be described with AsyncAPI, but yeah, is it a goal? do we want to claim that? that was never discussed

I think the basics in 3.0 release candidate are there already, operations at the root and stuff like that. So @Souvikns maybe do not hold but continue with 3.0 in focus only. The more people play with 3.0 at early stage, the better

PR #141, opened 2022-08-01
feat: add googlepubsub bindings

whitlockjc:
This commit contains two bindings, a `Channel Binding Object` and a `Message Binding Object` for Google Cloud Pub/Sub using the `googlepubsub` protocol name.

**Related issue(s)**
JSON Schema PR: https://github.com/asyncapi/spec-json-schemas/pull/253
Spec PR: https://github.com/asyncapi/spec/pull/836

### DIFF
--- a/googlepubsub/README.md
+++ b/googlepubsub/README.md
@@ -1,0 +1,170 @@
+# Google Cloud Pub/Sub Bindings
+
+This document defines how to describe Google Cloud Pub/Sub specific information with AsyncAPI.
+
+<a name="version"></a>
+
+## Version
+
+Current version is `0.1.0`.
+
+<a name="channel"></a>
+
+## Channel Binding Object
+
+The `Channel Bindings Object` is used to describe the Google Cloud Pub/Sub specific
+[Topic](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/create) details with AsyncAPI.
+
+Field Name | Type | Description
+---|---|---
+`bindingVersion`|String|The current version is `0.1.0`
+`labels`|Object|An object of key-value pairs _(These are used to categorize Cloud Resources like Cloud Pub/Sub Topics.)_
+`messageRetentionDuration`|String|Indicates the minimum duration to retain a message after it is published to the topic _(Must be a valid [Duration](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration).)_
+`messageStoragePolicy`|[Message Storage Policy Object](#message-storage-policy-object)|Policy constraining the set of Google Cloud Platform regions where messages published to the topic may be stored
+`schemaSettings`|[Schema Settings Object](#schema-settings-object)|Settings for validating messages published against a schema
+`topic`|String|The Google Cloud Pub/Sub Topic name
+
+<a name="message-storage-policy-object"></a>
+
+### Message Storage Policy Object
+
+The `Message Storage Policy Object` is used to describe the Google Cloud Pub/Sub
+[MessageStoragePolicy](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics#MessageStoragePolicy)
+Object with AsyncAPI.
+
+Field Name | Type | Description
+---|---|---
+`allowedPersistenceRegions`|String[]|A list of IDs of GCP regions where messages that are published to the topic may be persisted in storage
+
+<a name="schema-settings-object"></a>
+
+### Schema Settings Object
+
+The `Schema Settings Object` is used to describe the Google Cloud Pub/Sub
+[SchemaSettings](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics#SchemaSettings) Object with
+AsyncAPI.
+
+Field Name | Type | Description
+---|---|---
+`encoding`|String|The encoding of the message _(Must be one of the possible [Encoding](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics#encoding) values.)_
+`firstRevisionId`|String|The minimum _(inclusive)_ revision allowed for validating messages
+`lastRevisionId`|String|The maximum _(inclusive)_ revision allowed for validating messages
+`name`|String|The name of the schema that messages published should be validated against _(The format is `projects/{project}/schemas/{schema}`.)_
+
+<a name="channel-binding-example"></a>
+
+### Example
+
+```yaml
+# ...
+channels:
+  topic-avro-schema:
+    bindings:
+      googlepubsub:
+        topic: projects/your-project/topics/topic-avro-schema
+        schemaSettings:
+          encoding: json
+          name: projects/your-project/schemas/message-avro
+# ...
+  topic-proto-schema:
+    bindings:
+      googlepubsub:
+        topic: projects/your-project/topics/topic-proto-schema
+        messageRetentionDuration: 86400s
+        messageStoragePolicy:
+          allowedPersistenceRegions:
+          - us-central1
+          - us-central2
+          - us-east1
+          - us-east4
+          - us-east5
+          - us-east7
+          - us-south1
+          - us-west1
+          - us-west2
+          - us-west3
+          - us-west4
+        schemaSettings:
+          encoding: binary
+          name: projects/your-project/schemas/message-proto
+# ...
+```
+
+<a name="message"></a>
+
+## Message Binding Object
+
+The `Message Binding Object` is used to describe the Google Cloud Pub/Sub specific
+[PubsubMessage](https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage) details, alongside with pertintent
+parts of the Google Cloud Pub/Sub
+[Schema](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.schemas#Schema)
+Object, with AsyncAPI.
+
+Field Name | Type | Description
+---|---|---
+`bindingVersion`|String|The current version is `0.1.0`
+`attributes`|Object|Attributes for this message _(If this field is empty, the message must contain non-empty data. This can be used to filter messages on the subscription.)_
+`orderingKey`|String|If non-empty, identifies related messages for which publish order should be respected _(For more information, see [ordering messages](https://cloud.google.com/pubsub/docs/ordering).)_
+`schema`|[Schema Definition Object](#schema-definition-object)|Describes the schema used to validate the payload of this message
+
+<a name="schema-definition-object"></a>
+
+### Schema Definition Object
+
+The `Schema Definition Object` is used to describe the Google Cloud Pub/Sub
+[Schema]([Schema](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.schemas#Schema)) Object with AsyncAPI.
+While some of this information could be, or is, described using native AsyncAPI, for consistency it makes sense to
+provide this information here at all times, especially for cases where AsyncAPI does not natively support describing
+payloads using a supported Google Cloud Pub/Sub schema format like Protobuf.
+
+Field Name | Type | Description
+---|---|---
+`name`|String|The name of the schema
+`type`|String|The type of the schema
+
+<a name="message-binding-example"></a>
+
+### Example
+
+```yaml
+# ...
+components:
+  messages:
+    messageAvro:
+      bindings:
+        googlepubsub:
+          schema:
+            name: projects/your-project/schemas/message-avro
+            type: avro
+      contentType: application/json
+      name: MessageAvro
+      payload:
+        fields:
+        - name: message
+          type: string
+        name: Message
+        type: record
+      schemaFormat: application/vnd.apache.avro+yaml;version=1.9.0
+    messageProto:
+      bindings:
+        googlepubsub:
+          schema:
+            name: projects/your-project/schemas/message-proto
+            type: protobuf
+      contentType: application/octet-stream
+      name: MessageProto
+      payload: true
+# ...
+```
+
+<a name="operation"></a>
+
+## Operation Binding Object
+
+This object MUST NOT contain any properties. Its name is reserved for future use.
+
+<a name="server"></a>
+
+## Server Binding Object
+
+This object MUST NOT contain any properties. Its name is reserved for future use.
--- a/googlepubsub/json_schemas/channel.json
+++ b/googlepubsub/json_schemas/channel.json
@@ -1,0 +1,82 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://asyncapi.com/bindings/googlepubsub/channel.json",
+  "title": "Cloud Pub/Sub Channel Schema",
+  "description": "This object contains information about the channel representation for Google Cloud Pub/Sub.",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://raw.githubusercontent.com/asyncapi/spec-json-schemas/v2.14.0/schemas/2.4.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+    "bindingVersion": {
+      "type": "string",
+      "enum": [
+        "0.1.0"
+      ],
+      "description": "The version of this binding."
+    },
+    "labels": {
+      "type": "object"
+    },
+    "messageRetentionDuration": {
+      "type": "string"
+    },
+    "messageStoragePolicy": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "allowedPersistenceRegions": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "schemaSettings": {
+      "type": "object",
+      "additionalItems": false,
+      "properties": {
+        "encoding": {
+          "type": "string"
+        },
+        "firstRevisionId": {
+          "type": "string"
+        },
+        "lastRevisionId": {
+          "type": "string"
+        },
+        "name": {
+          "type": "string"
+        }
+      },
+      "required": ["encoding", "name"]
+    },
+    "topic": {
+      "type": "string"
+    }
+  },
+  "required": ["schemaSettings", "topic"],
+  "examples": [
+    {
+      "labels": {
+        "label1": "value1",
+        "label2": "value2"
+      },
+      "messageRetentionDuration": "86400s",
+      "messageStoragePolicy": {
+        "allowedPersistenceRegions": [
+          "us-central1",
+          "us-east1"
+        ]
+      },
+      "schemaSettings": {
+        "encoding": "json",
+        "name": "projects/your-project-id/schemas/your-schema"
+      }
+    }
+  ]
+}
--- a/googlepubsub/json_schemas/message.json
+++ b/googlepubsub/json_schemas/message.json
@@ -1,0 +1,55 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://asyncapi.com/bindings/googlepubsub/message.json",
+  "title": "Cloud Pub/Sub Channel Schema",
+  "description": "This object contains information about the message representation for Google Cloud Pub/Sub.",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://raw.githubusercontent.com/asyncapi/spec-json-schemas/v2.14.0/schemas/2.4.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+    "bindingVersion": {
+      "type": "string",
+      "enum": [
+        "0.1.0"
+      ],
+      "description": "The version of this binding."
+    },
+    "attributes": {
+      "type": "object"
+    },
+    "orderingKey": {
+      "type": "string"
+    },
+    "schema": {
+      "type": "object",
+      "additionalItems": false,
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "type": {
+          "type": "string"
+        }
+      },
+      "required": ["name", "type"]
+    }
+  },
+  "examples": [
+    {
+      "schema": {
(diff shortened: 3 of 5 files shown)


googlepubsub/README.md
```diff
@@ -0,0 +1,94 @@
+# Google Cloud Pub/Sub Bindings
+
+This document defines how to describe Google Cloud Pub/Sub specific information with AsyncAPI.
+
+<a name="version"></a>
+
+## Version
+
+Current version is `1.0.0`.
```
derberg, line 9: you want first version to be major?

googlepubsub/json_schemas/channel.json
```diff
@@ -0,0 +1,67 @@
+{
```
derberg, line 1: in both schemas, you're missing `bindingVersion` property. Please have a look at schemas from other bindings. You also do not have it in markdown definitions. It is necessary for user to have it, so they can specify what version of binding they're using. Bindings specs are versioned independently from main spec

googlepubsub/json_schemas/message.json
```diff
@@ -0,0 +1,48 @@
+{
```
derberg, line 1: I also noticed your schemas do not allow `patternProperties` with `x-` as the other schemas. Is it intentional, you do not want to allow extensions inside binding?

googlepubsub/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+Field Name | Type | Description
+---|---|---
+`attributes`|Object|An object of key-value pairs _(Attributes for this message.)_
+`orderingKey`|String|if non-empty, identifies related messages for which publish order should be respected
+`schema`|[Schema Definition Object](#schema-definition-object)|Describes the schema used to validate the payload of this message
+
+<a name="schema-definition-object"></a>
+
+### Schema Definition Object
```
derberg, line 70: This is quite a workaround 😄 
Shouldn't we try to fix it on a spec level rather than add such workaround gates on a binding level?

derberg (changes requested):
@whitlockjc I added few comments. Also general comment is that I'm missing examples in the markdown files, we should have them there too

Side note, to have it in the spec we will also need to add `googlepubsub` key to the spec in the list of bindings, but this is later, I can guide you and send some examples.

@fmvilas new binding on the way, please review

googlepubsub/README.md
whitlockjc: I wasn't sure what to use here, just not sure what else I see changing after we agree with its implementation.  I'm all ears for suggestions.

googlepubsub/json_schemas/channel.json
whitlockjc: My apologies, I'll update the PR to have these in both markdown and schemas.

googlepubsub/README.md
whitlockjc: Well, it's really not.  The whole purpose behind this is to describe the schema Cloud Pub/Sub uses to validate messages.  While AsyncAPI does have support for this already in the messages themselves, there is metadata on the schema object from a Cloud Pub/Sub perspective that is still useful for documenting the API.   Below is an example of how AsyncAPI's support for Avro and how its lack of Protobuf support would be documented:

```yaml
# ...
components:
    messages:
        messageAvroAsJSON:
            bindings:
                googlepubsub:
                    schema:
                        definition: |
                            {
                              "type" : "record",
                              "name" : "Message",
                              "fields" : [
                                {
                                  "name" : "message",
                                  "type" : "string"
                                }
                              ]
                            }
                        name: projects/jaydub-testing/schemas/message-avro
                        type: avro
            contentType: application/json
            name: projects/jaydub-testing/schemas/message-avro
            payload:
                fields:
                    - name: message
                      type: string
                name: Message
                type: record
            schemaFormat: application/vnd.apache.avro+yaml;version=1.9.0
        messageProto:
            bindings:
                googlepubsub:
                    schema:
                        definition: |
                            syntax = "proto2";

                            message Message {
                              required string message = 1;
                            }
                        name: projects/jaydub-testing/schemas/message-proto
                        type: protobuf
            contentType: application/octet-stream
            name: projects/jaydub-testing/schemas/message-proto
            payload: true
# ...
```

`name` is metadata that could/should be present, but I could see a case where if Protobuf were supported by AsyncAPI natively for Message payloads, maybe it wouldn't need to be present in the binding.  Thoughts?

googlepubsub/json_schemas/message.json
whitlockjc: It was intentional at first but I really can't think of a good reason why we shouldn't.  I'll add support.

whitlockjc:
I added examples to the Markdown file, and added `x-*` support to the JSON Schemas.

googlepubsub/README.md
lbroudoux: For other bindings, first public revision is usually marked as `0.1.0`. I think we can stick with convention.

googlepubsub/README.md
lbroudoux: I agree with @derberg: I think we absolutely have to avoid duplicating things in bindings when they should be defined in the message itself. Protobuf support should ideally follow the same approach as for Avro: having a dedicated `schemaFormat` entry for it. However due to `proto` syntax, we'll probably not be able to embed it in the AsyncAPI spec itself and use a `$ref` to external file.

The value you put in `name` (that should be human-readable according the spec) seems similar to the one put into `schemaSettings.name` at the Channel level. Is it just a duplication? Does it serve a specific purpose to have it structured like the schema name in your GCP project?

lbroudoux (commented):
Thanks for this PR! I made a few comments on version and Protobuf support.

googlepubsub/README.md
whitlockjc: I definitely want to stick with convention, so I'll update the PR and resolve this when it's done.

googlepubsub/README.md
whitlockjc: When it comes to embedding versus using a `$ref`, to me it's six in one hand and half another in the other. Ultimately, we want the API consumer to know as much as they can about the API contract and having a visual representation of unsupported schema types could serve some purpose. As for duplication, I'm all for removing it but if it means inconsistency _(duplication is removed for use case X but not use case Y)_, I personally would rather allow duplication and have consistency. Finally, for `name`, its structured is by design and a GCP concept.

fmvilas: Yeah, agree that's really a workaround. Let's wait until we have Protobuf support at the spec level using `schemaFormat`.

googlepubsub/README.md
```diff
@@ -0,0 +1,184 @@
...
+## Channel Binding Object
+
+The `Channel Bindings Object` is used to describe the Google Cloud Pub/Sub specific
+[Topic](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/create) details with AsyncAPI.
+
+Field Name | Type | Description
+---|---|---
+`bindingVersion`|String|The current version is `0.1.0`
+`labels`|Object|An object of key-value pairs _(These are used to categorize Cloud Resources like Cloud Pub/Sub Topics.)_
+`messageRetentionDuration`|String|Indicates the minimum duration to retain a message after it is published to the topic _(Must be a valid [Duration](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration).)_
```
fmvilas, line 22: I'm struggling to understand what a valid duration using a string format looks like. Can you provide an example of a full channel binding object below?

googlepubsub/README.md
```diff
@@ -0,0 +1,184 @@
...
+`messageStoragePolicy`|[Message Storage Policy Object](#message-storage-policy-object)|Policy constraining the set of Google Cloud Platform regions where messages published to the topic may be stored
```
fmvilas, line 23: This object only has a key for now. Do you think we can replace it with `allowedPersistenceRegions` instead? Or do you think this object might grow in the future?

googlepubsub/README.md
```diff
@@ -0,0 +1,184 @@
...
+The `Schema Settings Object` is used to describe the Google Cloud Pub/Sub
+[SchemaSettings](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics#SchemaSettings) Object with
+AsyncAPI.
+
+Field Name | Type | Description
+---|---|---
+`encoding`|String|The encoding of the message _(Must be one of the possible [Encoding](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics#encoding) values.)_
+`firstRevisionId`|String|The minimum _(inclusive)_ revision allowed for validating messages
+`lastRevisionId`|String|The maximum _(inclusive)_ revision allowed for validating messages
+`name`|String|The name of the schema that messages published should be validated against _(Format is `projects/{project}/schemas/{schema}`.)_
```
fmvilas, line 51:
```suggestion
`name`|String|The name of the schema that messages published should be validated against. The format is `projects/{project}/schemas/{schema}`.
```

googlepubsub/README.md
```diff
@@ -0,0 +1,184 @@
...
+[PubsubMessage](https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage) details, alongside with pertintent
+parts of the Google Cloud Pub/Sub
+[Schema](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.schemas#Schema)
+Object, with AsyncAPI.
+
+Field Name | Type | Description
+---|---|---
+`bindingVersion`|String|The current version is `0.1.0`
+`attributes`|Object|An object of key-value pairs _(Attributes for this message.)_
+`orderingKey`|String|if non-empty, identifies related messages for which publish order should be respected
```
fmvilas, line 103: Would you mind expanding on these two? I don't really understand what they do.

fmvilas (changes requested):

googlepubsub/README.md
whitlockjc: What do we do if/when GCP supports a new schema format? Do we disallow any sort of representation within the binding in hopes that AsyncAPI supports it? What do you think about making this optional instead of unnecessary?

googlepubsub/README.md
whitlockjc: _"The JSON representation for `Duration` is a `String` that ends in `s` to indicate seconds and is preceded by the number of seconds, with nanoseconds expressed as fractional seconds."_ So an example of one day is `86400s`.

googlepubsub/README.md
whitlockjc: I thought about this as well but since the Google API uses an object _(suggesting it it may grow in the future)_, I think it's safest to continue to model the Google API object structure.

googlepubsub/README.md
lbroudoux: Regarding `name`: I missed the point that I referred to `schemaSettings.name` and that it was a different thing than the topic name...

That said I realize that topic technical name was directly in the channel path. However this one should be human-readable, independent of binding (in case of multi-bindings supports) and may also container parameters!

In other bindings, we decided to add a specific property within the channel binding in order to add the technical name of the destination (look at `amqp` where we have `queue.name` or `exchange.name` ; we also have similar things in `ibmmq`). I think we should do the same here so that all the technical things related to a binding are kept into this binding.

We'd ended up with something like:

```yaml
channels:
  my-channel-name-with-business-meaning:
    bindings:
      googlepubsub:
        topic: projects/your-project/topics/topic-avro-schema
        schemaSettings:
          encoding: avro
          name: projects/your-project/schemas/message-avro
```

googlepubsub/json_schemas/channel.json
```diff
@@ -0,0 +1,79 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://asyncapi.com/bindings/amqp/channel.json",
```
lbroudoux, line 3: Change `$id` to `http://asyncapi.com/bindings/googlepubsub/channel.json`

googlepubsub/json_schemas/message.json
```diff
@@ -0,0 +1,60 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://asyncapi.com/bindings/amqp/channel.json",
```
lbroudoux, line 3: Change `$id` to `http://asyncapi.com/bindings/googlepubsub/message.json`

googlepubsub/json_schemas/channel.json
```diff
@@ -0,0 +1,79 @@
...
+  "title": "Cloud Pub/Sub Channel Schema",
+  "description": "This object contains information about the channel representation for Google Cloud Pub/Sub.",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/specificationExtension"
```
lbroudoux, line 10: Update to latest `https://raw.githubusercontent.com/asyncapi/spec-json-schemas/v2.14.0/schemas/2.4.0.json#/definitions/specificationExtension`

googlepubsub/json_schemas/message.json
```diff
@@ -0,0 +1,60 @@
...
+  "title": "Cloud Pub/Sub Channel Schema",
+  "description": "This object contains information about the message representation for Google Cloud Pub/Sub.",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/specificationExtension"
```
lbroudoux, line 10: Update to latest `https://raw.githubusercontent.com/asyncapi/spec-json-schemas/v2.14.0/schemas/2.4.0.json#/definitions/specificationExtension`

lbroudoux (changes requested):
Thanks for the update on PR!

I added a suggestion on moving topic name to specific channel binding property. I also had a more in-depth review and spotted some ids and refs that need to be updated in schemas.

googlepubsub/README.md
fmvilas: > What do we do if/when GCP supports a new schema format? Do we disallow any sort of representation within the binding in hopes that AsyncAPI supports it?

@whitlockjc that sounds like a perfect use case for an extension.

googlepubsub/README.md
fmvilas: Please add `duration` to the binding example, so it's easier to grasp.

googlepubsub/README.md
fmvilas: Please add `duration` to the binding example, so it's easier to grasp.

lbroudoux (changes requested):
Requested a few minor changes for updating $id and others.

googlepubsub/README.md
whitlockjc: I can go either way, but ultimately I need your agreement to get this submitted so I don't feel I've got much choice here. Requiring AsyncAPI to support a schema format formally versus allowing a binding to store potentially duplicate information might not be my first choice, the whole point of `x-*` is to fill the gaps so I don't care enough to push back.

whitlockjc: I updated the binding to add `topic` and to remove `definition` from `schemaSettings` per the discussion above.

googlepubsub/README.md
whitlockjc: Done, although we do have direct links to the source of truth describing this as well.

googlepubsub/json_schemas/message.json
whitlockjc: Done.

googlepubsub/json_schemas/channel.json
whitlockjc: Done.

googlepubsub/json_schemas/message.json
whitlockjc: Done.

whitlockjc:
All feedback has been addressed, please check if there are any places where I messed up the changes or if I missed anything.

whitlockjc:
I crossed the streams and was using `avro` and `protobuf` as valid values for `encoding` in the `SchemaSettings Object` when valid values are `binary`, `json` and `unknown` _(although we allow any value in there via JSON Schema right now by design to avoid having the currently-supported values hard-coded in for no real benefit)_.

googlepubsub/README.md
```diff
@@ -0,0 +1,170 @@
...
+components:
+  messages:
+    messageAvroAsJSON:
+      bindings:
+        googlepubsub:
+          schema:
+            name: projects/your-project/schemas/message-avro
+            type: avro
+      contentType: application/json
+      name: projects/your-project/schemas/message-avro
```
lbroudoux, line 140: Would you please change the `name` to something that doesn't look like a Google Cloud object name? Message name must be "machine-friendly" but having the same value as `bindings.googlepubsub.schema.name` can be confusing.

googlepubsub/README.md
```diff
@@ -0,0 +1,170 @@
...
+        type: record
+      schemaFormat: application/vnd.apache.avro+yaml;version=1.9.0
+    messageProto:
+      bindings:
+        googlepubsub:
+          schema:
+            name: projects/your-project/schemas/message-proto
+            type: protobuf
+      contentType: application/octet-stream
+      name: projects/your-project/schemas/message-proto
```
lbroudoux, line 155: Would you please change the `name` to something that doesn't look like a Google Cloud object name? Message name must be "machine-friendly" but having the same value as `bindings.googlepubsub.schema.name` can be confusing.

lbroudoux (changes requested):
Thanks for the update! This looks good to me. I just add a minor change request on `message.name` in examples - avoiding Google specific values that should now only live into `bindings.googlepubsub.schema.name`.

googlepubsub/README.md
whitlockjc: Done.  _(The examples were originally from a tool I wrote that generated AsyncAPI from Cloud Pub/Sub topology, wanted to make it clear where these values came from.)_

googlepubsub/README.md
whitlockjc: Done.  _(The examples were originally from a tool I wrote that generated AsyncAPI from Cloud Pub/Sub topology, wanted to make it clear where these values came from.)_

derberg (approved):
I think it looks good now 👍🏼 

sorry @whitlockjc on the pushback with schema stuff, but AsyncAPI is a mature project with a well-defined contribution guide and a stable release schedule. So I definitely encourage you to help us push forward proto support in the core spec. 

once others approve and we merge, I'll explain what you need to do in the main spec to enable this binding. Unless you already know?

fmvilas (approved):
LGTM 👍

fmvilas:
@char0n this PR can probably be included in 2.5.0. Pinging you so have it on your radar.

char0n:
@fmvilas yep, I'm already watching this. Thanks.

lbroudoux:
LGTM 👍  Thanks @whitlockjc for the successive updates.

whitlockjc:
> I think it looks good now 👍🏼
> 
> sorry @whitlockjc on the pushback with schema stuff, but AsyncAPI is a mature project with a well-defined contribution guide and a stable release schedule. So I definitely encourage you to help us push forward proto support in the core spec.
> 
> once others approve and we merge, I'll explain what you need to do in the main spec to enable this binding. Unless you already know?

No need to explain, I completely understand...although to my defense, I'm not sure where advocating for more documentation with the potential for duplication when AsyncAPI doesn't support a native schema format would jeopardize or question AsyncAPI being a _"mature project with a well-defined contribution guide and a stable release schedule."_ I think proto support would be awesome, something I'd love to help with. As for the next step, I _think_ I understand so I will get this sorted when the time comes.

derberg:
@whitlockjc hehe, no worries. I mean mature in the way that it is easy to improve, all is transparent, and if you drive things, you do not have to wait 2 years for the release or anything like that 😄 

In the case of features, yeah 🤷🏼 we have what the community wants, and believe me or not, proto is not something people talk about all the time 😄 But I definitely agree that support for proto and other formats, like xsd would be huge for AsyncAPI

whitlockjc:
We're on the same page, thank you so much for your help in getting this all sorted. I'm looking forward to working with AsyncAPI more in the future. ❤️

derberg:
ok then, looks like we can merge this one.

but as I wrote, merging doesn't mean it is released as part of the spec.

@whitlockjc if you want it in the September window, released with 2.5.0 then:
- you need open a PR to https://github.com/asyncapi/spec-json-schemas to add new binding key to JSON Schema
- you need to open a PR to https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md to add info about new binding 

@char0n is the release coordinator for the September release. So sync with him in https://github.com/asyncapi/spec/issues/830 as he just started and needs to prepare release branches. So definitely do not open PRs before he is done preparing them 😉 

all good?

char0n:
@derberg, given https://github.com/asyncapi/spec/issues/734, there's no need for ad-hoc release branches anymore. `next-spec` and `next-major-spec` are permanent release branches in my understanding.

derberg:
oh, @char0n I had no idea that @smoya already created `2.5.0` copy there https://github.com/asyncapi/spec-json-schemas/tree/next-spec/definitions/2.5.0. Normally you would have to make sure it happens 😄 

I still recommend updating these release branches with latest `master` as last time it was done in June

char0n:
> I still recommend updating these release branches with latest master as last time it was done in June

Yep, it's on my radar as next task.

whitlockjc:
I'll get on these other tasks tonight/tomorrow.

whitlockjc:
All necessary PRs have been created, all referencing the others.

char0n:
In order to merge this, we still require code review of one additional code owner. @KhudaDad414 can you please look into this PR? Thanks!

derberg:
@whitlockjc I invited you to this repo as maintainer. From now on, you are a CODEOWNER for googlepubsub binding. As a maintainer in one of AsyncAPI projects you also have right to join the AsyncAPI TSC and participate in decision making process for the entire initiative. Just lemme know if you are interested to learn more.

KhudaDad414 (approved):
LGTM!

char0n:
Good to go. Merging.

PR #150, opened 2022-09-26
feat: kafka bindings with support for scalar parameter values

vearutop:
**Description**

Based on [slack discussion](https://asyncapi.slack.com/archives/C34F2JV0U/p1663919937776929) this PR extends `kafka`  bindings to allow scalar values to optimize for a common case when key/clientId/groupId can be defined as constant.


kafka/README.md
```diff
@@ -35,8 +35,8 @@ This object contains information about the operation representation in Kafka.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="operationBindingObjectGroupId"></a>`groupId` | [Schema Object][schemaObject] | Id of the consumer group.
-<a name="operationBindingObjectClientId"></a>`clientId` | [Schema Object][schemaObject] | Id of the consumer inside a consumer group.
+<a name="operationBindingObjectGroupId"></a>`groupId` | [Schema Object][schemaObject] \| string | Id of the consumer group.
+<a name="operationBindingObjectClientId"></a>`clientId` | [Schema Object][schemaObject] \| string | Id of the consumer inside a consumer group.
```
fmvilas, line 39:
```suggestion
<a name="operationBindingObjectGroupId"></a>`groupId` | string \| [Schema Object][schemaObject] | Id of the consumer group or a [Schema Object][schemaObject] that validates it.
<a name="operationBindingObjectClientId"></a>`clientId` | [string \| [Schema Object][schemaObject] | Id of the consumer inside a consumer group or a [Schema Object][schemaObject] that validates it.
```

kafka/README.md
```diff
@@ -69,7 +69,7 @@ This object contains information about the message representation in Kafka.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="messageBindingObjectKey"></a>`key` | [Schema Object][schemaObject] \| [AVRO Schema Object](https://avro.apache.org/docs/current/spec.html) | The message key. **NOTE**: You can also use the [reference object](https://asyncapi.io/docs/specifications/v2.1.0#referenceObject) way. 
+<a name="messageBindingObjectKey"></a>`key` | [Schema Object][schemaObject] \| [AVRO Schema Object](https://avro.apache.org/docs/current/spec.html) \| string | The message key. **NOTE**: You can also use the [reference object](https://asyncapi.io/docs/specifications/v2.1.0#referenceObject) way.
```
fmvilas, line 72:
```suggestion
<a name="messageBindingObjectKey"></a>`key` | string \| [Schema Object][schemaObject] \| [AVRO Schema Object](https://avro.apache.org/docs/current/spec.html) | The message key or an object that validates the key. **NOTE**: You can also use a [Reference Object](https://asyncapi.io/docs/specifications/v2.1.0#referenceObject).
```

fmvilas (changes requested):
Left some comments. Some of the suggestions are not related to your modifications but I thought we could take the opportunity now.

Also, I'm missing that you bump the version of the document at the top. We don't have any automation there yet for bindings so you have to change it manually.

fmvilas (dismissed):
LGTM 👍 

Let's wait for @derberg, @lbroudoux, @KhudaDad414, and @dalelane reviews.

derberg:
as @lbroudoux and @dalelane are codeowners for kafka binding and @KhudaDad414 maintains JSON Schemas, I will not do the review here - already enough 👀 here 😃

fmvilas:
Oh yeah, forgot that Khuda maintains the schemas.

kafka/README.md
```diff
@@ -35,8 +35,8 @@ This object contains information about the operation representation in Kafka.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="operationBindingObjectGroupId"></a>`groupId` | [Schema Object][schemaObject] | Id of the consumer group.
-<a name="operationBindingObjectClientId"></a>`clientId` | [Schema Object][schemaObject] | Id of the consumer inside a consumer group.
+<a name="operationBindingObjectGroupId"></a>`groupId` | string \| [Schema Object][schemaObject] | Id of the consumer group or a [Schema Object][schemaObject] that validates it.
+<a name="operationBindingObjectClientId"></a>`clientId` | [string \| [Schema Object][schemaObject] | Id of the consumer inside a consumer group or a [Schema Object][schemaObject] that validates it.
```
KhudaDad414, line 39:
```suggestion
<a name="operationBindingObjectClientId"></a>`clientId` | string \| [Schema Object][schemaObject] | Id of the consumer inside a consumer group or a [Schema Object][schemaObject] that validates it.
```

KhudaDad414 (changes requested):
LGTM! just a little typo.

KhudaDad414 (approved):

kafka/README.md
```diff
@@ -69,7 +69,7 @@ This object contains information about the message representation in Kafka.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="messageBindingObjectKey"></a>`key` | [Schema Object][schemaObject] \| [AVRO Schema Object](https://avro.apache.org/docs/current/spec.html) | The message key. **NOTE**: You can also use the [reference object](https://asyncapi.io/docs/specifications/v2.1.0#referenceObject) way. 
+<a name="messageBindingObjectKey"></a>`key` | string \| [Schema Object][schemaObject] \| [AVRO Schema Object](https://avro.apache.org/docs/current/spec.html) | The message key or an object that validates the key. **NOTE**: You can also use a [Reference Object](https://asyncapi.io/docs/specifications/v2.1.0#referenceObject).
```
dalelane, line 72:
```suggestion
<a name="messageBindingObjectKey"></a>`key` | string \| [Schema Object][schemaObject] \| [AVRO Schema Object](https://avro.apache.org/docs/current/spec.html) | The message key or an object that validates the key. **NOTE**: You can also use a [Reference Object](https://asyncapi.io/docs/specifications/v2.4.0#referenceObject).
```

worth updating this while we're here?

dalelane (approved):
looks good - I think this will be helpful, thanks! 👍

kafka/README.md
lbroudoux: +1 for @dalelane suggestion

kafka/json_schemas/message.json
```diff
@@ -11,13 +11,22 @@
...
     "key": {
-      "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/schema",
+      "oneOf": [
+        {
+          "description": "Constant value of key.",
+          "type": "string"
+        },
+        {
+          "description": "Schema that describes the allowed values of key.",
+          "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/schema"
```
lbroudoux, line 21: Should be updated to `https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.14.0/schemas/2.4.0.json#/definitions/schema`

lbroudoux (changes requested):
LGTM. Just update schema references in `message.json` and we're good to go.

vearutop:
Oh, sorry for delay, I see these bindings were recently versioned with 0.3.0 for other changes, should I bump version to 0.3.1 (or to 0.4.0) here?

fmvilas:
@vearutop Yeah, bump it to 0.4.0 as it's actually a new "feature".

dalelane (dismissed):

fmvilas (approved):
LGTM 👍

derberg:
sorry folks, I wrote I won't review, but actually, today during 3.0 spec meeting it popped up and I read the thread on slack because of it.

can you help me understand why it is better to allow just plain string instead of following schema object:
```
          groupId:
            const: my-group
```

isn't it only a problem of bad example that suggest `enum` while better to have `const`?

vearutop:
Good thing about scalar value is that it has clear intent and is hard to misuse. When I first saw that `groupId` accepts schema I was genuinely confused (by that time I haven't checked examples yet). 

So in my opinion schema covers advanced cases where user would have to look in the docs anyway, while "string" simplifies basic case (constant value). It also works a bit like documentation, suggesting the type of value (schema does not limit itself in any way and lends itself for misuse if user is not careful about what goes into schema).

fmvilas:
Yeah, also a plain string is always a scalar value. A Schema may not be possible to convert down to a scalar value if it doesn't have `const` or an `enum` with a single element array. They serve two different purposes. A producer may want to know which exact value to send. A consumer may want to know how to validate it.

For instance, a publisher may send `key` as `user.signedupV1` and another publisher may set it to `user.signedupV2` but a potential consumer may just want to check that the value is either `user.signedupV1` or `user.signedupV2`, or that it simply matches a specific regex pattern.

The publishers want to have the scalar value they send. The consumer wants to have a Schema object that validates it. Does it make sense?

dalelane:
@lbroudoux do you think that covers everything you saw?

derberg:
yeah, I still think better is:

```yaml
groupId:
    const: my-group
    description: You can always specify more details for the user if you want
```

than:
```yaml
groupId: my-group
```

anyway, I suggest you provide an example for string and probably adjust example with single element `enum` to `const`

lbroudoux:
Still looking good to me. 👍 

Also I think it could be nice to illustrate both cases through different examples like:


```yaml
channels:
  test:
    publish:
      message:
        bindings:
          kafka:
            key: 'user.signedupV1'
```

OR

```yaml
channels:
  test:
    publish:
      message:
        bindings:
          kafka:
            key:
              type: string
              enum: ['user.signedupV1', 'user.signedupV2']
```

IMO this clearly illustrates that you can use this notation shortcut for self-explanatory values but have to switch to schema notation once you handle more complex cases.  In that later cases, additional `description` can make sense to explain the range of possible values.

lbroudoux (approved):

vearutop:
Hi, is there anything I can do to improve this PR?

fmvilas:
@vearutop Yeah, I think it's just a matter of adding examples as @derberg and @lbroudoux suggested.

Issue #153, opened 2022-10-06
Typo in JSON schema for Solace server binding 


195858:
#### Describe the bug
Typo in JSON schema for Solace server binding 

#### How to Reproduce
Review the JSON schema - it declares a property "msvVpn", the example given in the schema uses "msgVpn" which is correct

#### Expected behavior
The property should be called "msgVpn"

PR #154, opened 2022-10-06
fix: solace bindings - typo "msvVpn" should be "msgVpn"

195858:
https://github.com/asyncapi/bindings/issues/153

**Description**

Fixed typo in Solace server binding JSON schema

**Related issue(s)**
Resolves
https://github.com/asyncapi/bindings/issues/153

### DIFF
--- a/solace/json_schemas/server.json
+++ b/solace/json_schemas/server.json
@@ -11,7 +11,7 @@
     }
   },
   "properties": {
-    "msvVpn": {
+    "msgVpn": {
       "type": "string",
       "description": "The name of the Virtual Private Network to connect to on the Solace broker."
     },
--- a/solace/json_schemas/server.json
+++ b/solace/json_schemas/server.json
@@ -26,7 +26,7 @@
   "examples": [
     {
       "msgVpn": "ProdVPN",
-      "bindingVersion": "0.2.0"
+      "bindingVersion": "0.2.1"
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


195858:
2nd commit fixes this issue 

https://github.com/asyncapi/bindings/issues/155

damaru-inc (dismissed):

solace/json_schemas/server.json
```diff
@@ -11,7 +11,7 @@
     }
   },
   "properties": {
-    "msvVpn": {
```
KhudaDad414, line 14: Hey @195858, It would be great if you could increment the binding version as well.

195858: Bumped the version number to 0.2.1

damaru-inc (approved):

char0n:
Hi everybody,

I was wondering why are we increasing the binding version to `0.2.1`. As the markdown file is the source of truth for the binding, nothing has been changed in that file. This PR only addresses the fix in accompanied JSON Schema which is only supplement of the spec.

VisualBean:
> Hi everybody,
> 
> I was wondering why are we increasing the binding version to `0.2.1`. As the markdown file is the source of truth for the binding, nothing has been changed in that file. This PR only addresses the fix in accompanied JSON Schema which is only supplement of the spec.

I agree, if it doesn't change the actual spec, then the version should not change. (Also renaming a property is a breaking change, so it would deem a major version increment, not a patch version increment).

PR #156, opened 2022-10-10
chore: add @smoya, @dalelane, and @char0n as code owners

fmvilas:
## Description

Adding @smoya and @char0n as code owners. Also adding @dalelane who seemed to be missing. Even though he's on the kafka binding but as a spec code owner I think he should also be code owner of all the bindings. What do you think?

## Related issues
https://github.com/asyncapi/spec/pull/846
https://github.com/asyncapi/spec-json-schemas/pull/271

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,7 +6,7 @@
 
 # The default owners are automatically added as reviewers when you open a pull request unless different owners are specified in the file.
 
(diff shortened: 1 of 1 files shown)


derberg (approved):
makes sense to me

dalelane (approved):

PR #143, opened 2022-08-18
docs: make usage of Schema Object consistent

char0n:
**Description**

[Main specification document](https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md) says this:

image

All the Schema Objects in main specification document has following type notation:

image

All binding objects specification documents contain following type notation:

image

---

From this, it is apparent that type notation for fields using `Schema Object`, is not consistent within the spec and probably needs to be clarified. There is no reason (at least none that I can see currently) why to constrain binding objects fields to using `Schema Object` and not allow `Reference Object` in it's place as well. The AsynAPI specification tells us following: `Alternatively, any time a Schema Object can be used, a Reference Object can be used in its place.` - this can also be interpreted that using `Reference Object` in binding objects fields is allowed, but if we check the [schemas](https://github.com/asyncapi/bindings/blob/master/http/json_schemas/message.json#L15), they tell us `Reference Object` is not allowed. As we already [stipulated before](https://github.com/asyncapi/spec/pull/781), the specification is the source of truth and not the schemas.

This PR open possibility to use `Reference Object` in binding objects fields and updates schemas to use latest versions of AsyncAPI  schemas (2.4.0). This also enhanced re-usability aspect of the spec. 

**Related issue(s)**

https://github.com/swagger-api/apidom/issues/1832

PR #157, opened 2022-10-11
feat: session expiry interval to mqtt5 bindings

sudoshreyansh:
**Description**

Adds `sessionExpiryInterval` to MQTT5 server bindings, which is required for https://github.com/asyncapi/glee/pull/323

### DIFF
--- a/mqtt5/README.md
+++ b/mqtt5/README.md
@@ -13,9 +13,27 @@ Current version is `0.1.0`.
 
 ## Server Binding Object
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
+This object contains information about the server representation in MQTT5.
 
+##### Fixed Fields
 
+Field Name | Type | Description
+---|:---:|---
+<a name="serverBindingObjectSessionExpiryInterval"></a>`sessionExpiryInterval` | integer | Session Expiry Interval in seconds.
+<a name="serverBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
+
+This object MUST contain only the properties defined above.
+
+##### Example
+
+```yaml
+servers:
+  production:
+    bindings:
+      mqtt5:
+        sessionExpiryInterval: 60
+        bindingVersion: 0.1.0
+```
 
 <a name="channel"></a>
 
--- a/mqtt5/json_schemas/server.json
+++ b/mqtt5/json_schemas/server.json
@@ -1,0 +1,34 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://asyncapi.com/bindings/mqtt5/server.json",
+  "title": "Server Schema",
+  "description": "This object contains information about the server representation in MQTT5.",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+
+    "sessionExpiryInterval": {
+      "type": "integer",
+      "description": "Session Expiry Interval in seconds."
+    },
+    "bindingVersion": {
+      "type": "string",
+      "enum": [
+        "0.1.0"
+      ],
+      "description": "The version of this binding. If omitted, 'latest' MUST be assumed."
+    }
+  },
+  "examples": [
+    {
+      "sessionExpiryInterval": 60,
+      "bindingVersion": "0.1.0"
+    }
+  ]
+}
+  


sudoshreyansh:
@fmvilas updated the binding version. Can you please re-review?

mqtt5/json_schemas/server.json
```diff
@@ -0,0 +1,34 @@
...
+  "description": "This object contains information about the server representation in MQTT5.",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^x-[\\w\\d\\.\\-\\_]+$": {
+      "$ref": "https://raw.githubusercontent.com/asyncapi/asyncapi-node/v2.7.7/schemas/2.0.0.json#/definitions/specificationExtension"
+    }
+  },
+  "properties": {
+
```
fmvilas, line 14:
```suggestion
```

fmvilas (approved):
LGTM 👍 Let's wait for others to review too.

Pinging @KhudaDad414 @derberg @dalelane @smoya @char0n

dalelane (commented):
@rcoppen is better placed to comment on this than me, as my MQTT knowledge is pretty rusty... but I'll ask some ignorant questions anyway :-) 

Is session expiry an attribute of the server? I thought it was something the client could choose, with different clients connecting to the same server allowed to choose their own (and different) expiries?

rcoppen:
I agree with @dalelane, is the intention to define a server configuration that always enforces a single Session Expiry for every client connection?

smoya:
I agree @dalelane and @rcoppen. Session expiration is only a client decision. I tried to think on some way of adding this to the binding, like setting some "max" session TTL, but still it doesn't fit IMHO.
Regarding Glee, you could still use [extensions](https://www.asyncapi.com/docs/reference/specification/v2.5.0#specificationExtensions).

rcoppen:
Looking at the `mqtt` binding a similar problem exists for `clientId`. If a value is presented in the AsyncAPI doc then should every client connect with that `clientId`? This would limit the server to one active client per doc instance. One approach would be to assert a server default in the binding spec should the client not present value. For `clientId` (assuming MQTT 3.1.1 or 5.0) then that could be a default of "not present" and the server would then create one.

smoya:
> Looking at the `mqtt` binding a similar problem exists for `clientId`. If a value is presented in the AsyncAPI doc then should every client connect with that `clientId`? This would limit the server to one active client per doc instance. 

Good catch! 

> One approach would be to assert a server default in the binding spec should the client not present value. For `clientId` (assuming MQTT 3.1.1 or 5.0) then that could be a default of "not present" and the server would then create one.

I wouldn't either set a default, since then the same issue you described earlier could happen.
I would suggest that the type becomes a `Schema` instead, so users can configure the schema the value should fulfill, including things like default value, regex, etc etc.

But that's for another Issue/PR. Are you happy opening a new issue with this @rcoppen ? I can do otherwise.

rcoppen:
@smoya np, will open an issue for `clientId`. My initial thinking was that since MQTT 3.1.1 states that the server will autogenerate a clientId if the client doesn't provide one, then that could be one way to overcome the problem. However I agree that's a discussion for the new issue.

fmvilas:
Yeah, these things need to be clarified. Y'all are making a good point that this is something the client should decide, maybe even on runtime. I think it could be a good idea to follow the direction of Kafka and WS bindings, and allow either a scalar value or a Schema Object value. The Schema Object should give the client enough flexibility to calculate things on runtime but still make sure it follows a specific pattern or limits. The scalar value would be great for those cases in which we want to enforce a specific value. For instance, when generating code.

sudoshreyansh:
I agree with @fmvilas. I have made the required changes.

mqtt5/README.md
```diff
@@ -6,16 +6,34 @@ This document defines how to describe MQTT 5-specific information on AsyncAPI.
...
 ## Server Binding Object
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
+This object contains information about the server representation in MQTT5.
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---:|---
+<a name="serverBindingObjectSessionExpiryInterval"></a>`sessionExpiryInterval` | [Schema Object][schemaObject] \| integer | Session Expiry Interval in seconds.
```
fmvilas, line 22: > Session Expiry Interval in seconds.

@sudoshreyansh if the value is a Schema Object, then the description above is not true. Please adjust.

mqtt5/README.md
```diff
@@ -6,16 +6,34 @@ This document defines how to describe MQTT 5-specific information on AsyncAPI.
...
+This object MUST contain only the properties defined above.
 
+##### Example
+
+```yaml
+servers:
+  production:
+    bindings:
+      mqtt5:
+        sessionExpiryInterval: 60
```
fmvilas, line 34: Would be cool to show an example using Schema Object too.

mqtt5/json_schemas/server.json
```diff
@@ -0,0 +1,42 @@
...
+      "oneOf": [
+        {
+          "type": "integer",
+          "minimum": 0
+        },
+        {
+          "$ref": "https://raw.githubusercontent.com/asyncapi/spec-json-schemas/v2.14.0/schemas/2.4.0.json#/definitions/schema"
+        }
+      ],
+      "description": "Session Expiry Interval in seconds."
```
fmvilas, line 25: Same as above. Something to tweak for the Schema Object case.

fmvilas (changes requested):

mqtt5/README.md
sudoshreyansh: @fmvilas I have updated the description.

mqtt5/README.md
sudoshreyansh: @fmvilas added another example using Schema Object.

dalelane (approved):

fmvilas:
@KhudaDad414 can you have a look as well, please? 🙏

KhudaDad414 (approved):
LGTM!

mqtt5/README.md
```diff
@@ -6,16 +6,44 @@ This document defines how to describe MQTT 5-specific information on AsyncAPI.
...
-This object MUST NOT contain any properties. Its name is reserved for future use.
-
-
+This object contains information about the server representation in MQTT5.
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---:|---
+<a name="serverBindingObjectSessionExpiryInterval"></a>`sessionExpiryInterval` | [Schema Object][schemaObject] \| integer | Session Expiry Interval in seconds or a Schema Object containing the definition of the interval.
```
char0n, line 22:
```suggestion
<a name="serverBindingObjectSessionExpiryInterval"></a>`sessionExpiryInterval` | [Schema Object][schemaObject] \| [Reference Object](referenceObject) \| integer | Session Expiry Interval in seconds or a Schema Object containing the definition of the interval.
```

mqtt5/README.md
```diff
@@ -38,4 +66,6 @@ This object MUST NOT contain any properties. Its name is reserved for future use
 
 ## Message Binding Object
 
-This object MUST NOT contain any properties. Its name is reserved for future use.
\ No newline at end of file
+This object MUST NOT contain any properties. Its name is reserved for future use.
+
+[schemaObject]: https://www.asyncapi.com/docs/specifications/2.4.0/#schemaObject
```
char0n, line 71:
```suggestion
[schemaObject]: https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#schemaObject
[referenceObject]: https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#referenceObject
```

mqtt5/json_schemas/server.json
```diff
@@ -0,0 +1,49 @@
...
+  "properties": {
+
+    "sessionExpiryInterval": {
+      "oneOf": [
+        {
+          "type": "integer",
+          "minimum": 0
+        },
+        {
+          "$ref": "https://raw.githubusercontent.com/asyncapi/spec-json-schemas/v2.14.0/schemas/2.4.0.json#/definitions/schema"
```
char0n, line 22: Can we replace the `$ref` with following two objects?

```json
   {
    "$ref": "https://asyncapi.com/definitions/2.4.0/schema.json"
   },
   {
     "$ref": "https://asyncapi.com/definitions/2.4.0/Reference.json"
   }
```

char0n (commented):
LGTM. I've suggested couple of changes, due to https://github.com/asyncapi/bindings/pull/143

sudoshreyansh:
I have made the changes @char0n 
Could you please review?

dalelane (approved):

fmvilas (approved):

fmvilas:
@KhudaDad414 would you mind reapproving if you agree with the changes?

KhudaDad414 (approved):
LGTM!

PR #169, opened 2022-11-09
fix(ibmmq): make topic field required

char0n:
I'm not sure if my assumption is correct here, but logically I would expect this field to be **REQUIRED** when `destinationType=topic`

image

Just as the `queue` is **REQUIRED** when `destinationType=queue`. 

image

I might be mistaken, but issued this PR to open a discussion around it. As a tooling author I'm seeking full understanding in purpose of validation.

IMHO the schemas are also saying the same as me - https://github.com/asyncapi/bindings/blob/master/ibmmq/json_schemas/channel.json#L88


char0n:
@rcoppen thanks for the explanation, it makes sense.

So probably we should change the [schema](https://github.com/asyncapi/bindings/blob/master/ibmmq/json_schemas/channel.json#L88) then - correct me if I'm wrong, the schema currently says the following: 

**IF** `destinationType=topic` **THEN** `topic` fixed field is **REQUIRED**

What we probably want it to say is:

**IF** `destinationType=topic` **THEN** `topic` fixed field **MAY** be present

rcoppen:
@char0n thanks for progressing this.
Is there a house rule on RFC2119 *MAY* clauses? Introducing optionality has the drawback of proliferating multiple ways of defining the same Topic Name making it potentially more difficult to implement something. On the other hand, forcing the AsyncAPI channel to be overridden (potentially with the same value) could leave room for error.

rcoppen:
@char0n do we want to go forward with the change to make `topic` REQUIRED or leave it as OPTIONAL?

rcoppen:
@char0n, yes that works.

Thanks


Richard

char0n:
This PR is succeeded by https://github.com/asyncapi/bindings/pull/176

PR #171, opened 2022-11-10
feat: add topicConfiguration property to channel object

### LINKED ISSUE #163: add `topicConfiguration` property to kafka channel binding
#### Reason/Context
Many topic configuration properties influence how API consumers can use a topic that is owned by an API provider. To transparently communicate relevant topic configurations, it should be part of the AsyncAPI spec.

Example:
If the `cleanup.policy` and the related configurations [`delete.retention.ms`](https://kafka.apache.org/documentation/#topicconfigs_delete.retention.ms), [`retention.ms`](https://kafka.apache.org/documentation/#topicconfigs_retention.ms), and [`retention.bytes`](https://kafka.apache.org/documentation/#topicconfigs_retention.bytes) were available in the AsyncAPI spec, an API consumer could at least conduct the following:
- Longest possible time to recover from a failure
- That messages can get lost due to compaction.

#### Description
Add a new `topicConfiguration` property to the kafka channel binding. API providers can add relevant configuration properties to the API-spec.

```yaml
topicConfiguration:
  description: Topic configuration properties that are relevant to the consumer.
  type: object
  additionalProperties: true  
```
By adding a generic `topicConfiguration` property, the kafka binding stays generic to configuration evolution in kafka itself.

thake:
resolves #163 by introducing a new `topicConfiguration`property to the channel object.


kafka/README.md
```diff
@@ -65,7 +67,38 @@ channels:
...
+        bindingVersion: '0.4.0'
+```
+<a name="topicConfiguration"></a>
+## TopicConfiguration Object
+
+This objects contains information about the API relevant topic configuration in Kafka.
+
+Field Name |  Type   |                                                                             Description                                                                              | Applicability [default] | Constraints
+---|:-------:|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------:|:-----------------------:|---
+<a name="topicConfigurationCleanupPolicy"></a>`cleanup.policy` |  array  |                            The [`cleanup.policy`](https://kafka.apache.org/documentation/#topicconfigs_cleanup.policy) configuration option.                          |        OPTIONAL         | array may only contain `delete` and/or `compact`
```
dalelane, line 85: I think the other four config properties make total sense to include, as they are properties that can cause problems for an application if they don't behave in a compatible way. 

But I'm not as sure about this one. Why do you think client applications need to know about the cleanup policy?

kafka/README.md
```diff
@@ -65,7 +67,38 @@ channels:
...
+<a name="topicConfigurationRetentionMs"></a>`retention.ms` | integer |                               The [`retention.ms`](https://kafka.apache.org/documentation/#topicconfigs_retention.ms) configuration option.                                    |        OPTIONAL         | Must at least be -1
```
dalelane, line 86: I know what you mean by the "Must at least be -1" constraint, but I think it's possibly a little confusing.

Perhaps we could just defer to the official Kafka docs?

Or if we need to specify something here, we could say something like:
> If set, this should be equal to `-1` to indicate that no time limit is applied, otherwise it should be a value greater than 0.

kafka/README.md
```diff
@@ -65,7 +67,38 @@ channels:
...
+<a name="topicConfigurationRetentionBytes"></a>`retention.bytes` | integer |                       The [`retention.bytes`](https://kafka.apache.org/documentation/#topicconfigs_retention.bytes) configuration option.                                                           |        OPTIONAL         | MUST at least be -1
```
dalelane, line 87: as above, I think the constraint wording could be misleading

Perhaps we could just defer to the official Kafka docs?

Or if we need to specify something here, we could say something like:
> If set, this should be equal to `-1` to indicate that no size limit is applied, otherwise it should be a value greater than 0.

kafka/README.md
```diff
@@ -65,7 +67,38 @@ channels:
...
+<a name="topicConfigurationDeleteRetentionBytes"></a>`delete.retention.ms` | integer |             The [`delete.retention.ms`](https://kafka.apache.org/documentation/#topicconfigs_delete.retention.ms) configuration option.                                               |        OPTIONAL         | MUST at least be 0
```
dalelane, line 88: as above, I think "at least" is possibly confusing. Not sure what the best way to phrase this, but essentially I'm trying to say that something like "If set, this MUST be 0 or a positive integer"

kafka/README.md
```diff
@@ -65,7 +67,38 @@ channels:
...
+<a name="topicConfigurationMaxMessageBytes"></a>`max.message.bytes` | integer |                    The [`max.message.bytes`](https://kafka.apache.org/documentation/#topicconfigs_max.message.bytes) configuration option.                                      |        OPTIONAL         | MUST at least be 0
```
dalelane, line 89: as above

dalelane (commented):
overall, I really like this - I think it's going to be super helpful. 

I have a few minor nit-picks over how we word the constraints, but just trying to make sure we phrase it in a way that is easy for anyone to understand.

My only real concern is about the inclusion of cleanup.policy - I'm not sure whether clients really need to know about this one.

kafka/README.md
thake: The cleanup policy defines if a client can expect to receive every record written to the topic. 

`compact`: A client can expect to receive the last record for a specific key, no matter if it has been written yesterday of 2 years ago. A client has to cope with the fact that he may only get the last record for that specific key.

`delete`: A client can expect to get all records written before `retension.ms` or `retension.bytes`. The client can expect to get all records in that timeframe independent of the record key.

This information is important for the design of applications consuming topics

kafka/README.md
thake: I like the idea of just referring to the kafka docs :) less duplication and always correct!

thake:
@dalelane thanks for your fast and very good review :+1:  I've updated the README.md to always point to the kafka documentation for constraints.

dalelane (approved):
LGTM - thanks!

@lbroudoux - what do you think?

KhudaDad414 (approved):
LGTM!

lbroudoux (approved):
LGTM. Love your proposition!!

lbroudoux:
> Is there anything left to do here? Are any more approvals needed? I'm unsure of the proceedings 😏

@thake Sorry for the delay and thanks for your PR. I really like your proposition: it adds useful complements!

KhudaDad414 (approved):
🚢 🚢

lbroudoux:
@dalelane @KhudaDad414  All checks are green. I think we're good for merging! Let me know so that I can proceed.

Issue #113, opened 2022-01-23
Move binding JSON schema files to main JSON schema repository


jonaslagoni:
#### Reason/Context
Currently, we store the binding JSON Schema files under each binding. However, it makes them hard to reach for others. Especially if we wish to provide them with our JSON Schema for the spec - https://github.com/asyncapi/spec-json-schemas/tree/master/schemas

See https://github.com/asyncapi/spec/issues/507

#### Description
I suggest we move all the JSON Schema files for bindings to https://github.com/asyncapi/spec-json-schemas. That way we match the current setup for the specification and make it easy to reference them directly in the spec JSON Schema https://github.com/asyncapi/spec-json-schemas/blob/abeac12417603de5b5fd57be4af4c8289e274a2b/schemas/2.2.0.json#L773

dalelane:
Makes sense to me to keep it together. I'd be wary of deleting them from the bindings repo though, in case people are using URLs like https://raw.githubusercontent.com/asyncapi/bindings/master/kafka/json_schemas/message.json as `$ref` values - as we don't want to break anything. 

Maybe a README in the json_schemas folders in the bindings repo explaining that the copies there are no longer being updated, and pointing at the correct new location? 

I realise I'm being uber-cautious and conservative here! I certainly can't find any use of `https://raw.githubusercontent.com/asyncapi/bindings/master/` from a quick search of public github repos, so I'm open to being told that I'm being over paranoid :-)

derberg:
I opted in keeping them here when they were added to make contribution process easier -> one repo to push changes.

but we already have json schema separated from `spec` repo for a good reason, to easily bug fix errors in json schema without releasing the spec. So yeah, I think the best would be to move JSON Schemas from here to `spec-json-schemas`.

we just need to have it clear in readme and have clear guidelines for maintainers from this repo, to accept changes only if PR is opened against `spec-json-schemas` too

jonaslagoni:
I get the cautiousness @dalelane, I didn't really think of that use case as there was no library to enable the usage I automatically dismissed they were being used 😆 

We could just reference the files in `spec-json-schemas` using `https://raw.githubusercontent.com/asyncapi/bindings/master/kafka/json_schemas/message.json`? The bundler that is in place for https://github.com/asyncapi/spec-json-schemas/pull/128 will automatically fetch the binding schemas and bundle them into the main schema. So there is no problem there if that's what we want to do instead.

magicmatatjahu:
[...]

PR #179, opened 2023-01-17
feat: enable extensions in bindings

derberg:
Related https://github.com/asyncapi/spec/pull/893

This PR once and for all makes it clear that bindings are also extendable like other parts of AsyncAPI spec with specification extensions (`x-whatever`)

- this PR updates all bindings with info about specification extension
- this PR bumps versions of bindings and update their schemas

---

## One open questions, something maybe controversial

In this PR I also suggest that text:
```
This object MUST NOT contain any properties. Its name is reserved for future use.
```

is replaced everywhere with:
```
This object does not contain any fixed fields. Its name is reserved for future use.

This object MAY be extended with [Specification Extensions](https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#specification-extensions).
```

The fact that we do not have any fixed fields for some binding should not block users from experimenting with extensions. So above change is needed if we want to truly enable extensions in all the bindings.
This change is visible in [this commit](https://github.com/derberg/bindings/commit/5d19c73b921d53b3cee1325ea01280d4051ae695). I did not provide schemas for these bindings yet, as first I want to be sure it is widely accepted.


VisualBean (commented):
Why are the versions changing for the Json schemas, but nothing actually changed in them?
I believe most, if not all, already had the extension regex.

derberg:
@VisualBean yes, you are right, we had extensions enabled on JSON Schema level. But readme text was saying something different. JSON Schema is considered to be a "tool" helpful for validation and other stuff. Individual binding readme is what is the "truth", so if we make some changes there that change behaviour, we need to bump the version of the binding. And yeah, texts in readmes were saying that basically "no props" should be added (which also includes extension).

Does that make sense?

VisualBean:
> @VisualBean yes, you are right, we had extensions enabled on JSON Schema level. But readme text was saying something different. JSON Schema is considered to be a "tool" helpful for validation and other stuff. Individual binding readme is what is the "truth", so if we make some changes there that change behaviour, we need to bump the version of the binding. And yeah, texts in readmes were saying that basically "no props" should be added (which also includes extension).
> 
> Does that make sense?

I'm surprised that the "text" side of the documentation is what is considered the "truth" - with that in mind, the change makes sense.

VisualBean (approved):

derberg:
@VisualBean yes, mainly because not all the things can be described with JSON Schema.

It is the same with main spec. 
- this is source of truth https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md
- this is just a helper tool https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/2.5.0.json

this is why also binding's JSON Schema files will be also moved to `spec-json-schemas` repo -> https://github.com/asyncapi/bindings/issues/113

---

Thanks so much for the approval and quick reaction 🙇

VisualBean:
> @VisualBean yes, mainly because not all the things can be described with JSON Schema.
> 
> It is the same with main spec. 
> - this is source of truth https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md
> - this is just a helper tool https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/2.5.0.json
> 
> this is why also binding's JSON Schema files will be also moved to `spec-json-schemas` repo -> https://github.com/asyncapi/bindings/issues/113
> 
> ---
> 
> Thanks so much for the approval and quick reaction 🙇 

Any time. I think my reason for thinking it was the schema that was the truth, is that it is machine verifiable, so to speak, where the "text" isn't. 

Anyway, you are very welcome.

KhudaDad414 (approved):
LGTM!

CameronRushton (approved):
Solace stuff looks good to me. Thanks!

GeraldLoeffler (approved):
reviewed and approved anypointmq binding

dalelane (approved):
makes sense to me 👍

rcoppen (approved):

char0n (approved):
> The fact that we do not have any fixed fields for some binding should not block users from experimenting with extensions. So above change is needed if we want to truly enable extensions in all the bindings.

Amen ;]


LGTM and nice work! :tada: 

The sentence `This object MAY be extended with [Specification Extensions](https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#specification-extensions)` is always in right place, right under the list of fixed fields (if they exists).

derberg:
only few of you left to accept 😄 

@whitlockjc @fmvilas @smoya @lbroudoux @damaru-inc

whitlockjc (approved):
👍 for `googlepubsub`.

damaru-inc (approved):

derberg:
@lbroudoux @fmvilas @smoya need your approval folks

smoya (approved):
LGTM! 🚀🌔

lbroudoux:
Looks good to me 👍

lbroudoux (approved):
Kafka part looks good 🚀

fmvilas:
@derberg I think we can merge this now, right?

PR #188, opened 2023-02-25
docs: removed subscribe for operation type - retain

### LINKED ISSUE #187: MQTT `retain` flag should be applied only to `publish` operations
#### Reason/Context
The `retain` flag in [MQTT binding](https://github.com/asyncapi/bindings/blob/master/mqtt/README.md?plain=1#L72) should only be applied when the type of operation is `publish`. It can be applied to publish and subscribe according to the current bindings.

#### Solution
Remove `subscribe` from the `Applies to` column in the operation binding object.

kushalnl7:
**Description**
Removed `subscribe` from the `Applies to` column in the operation binding object.

**Related issue(s)**
Resolves #187


jonaslagoni (changes requested):
Remember to update the version of the binding as well to `0.2.0`, just to note, this is a breaking change 🙂

char0n:
@kushalnl7 do you need any assistance to push this PR forward? As @jonaslagoni mentioned in https://github.com/asyncapi/bindings/pull/188#pullrequestreview-1645692368, we need to bump the binding version as we changed it's behavior by this change.

derberg:
no follow ups, closing

Issue #182, opened 2023-02-02
Inspecting bindings for spec 3.0


jonaslagoni:
#### Reason/Context
We have no overview of how the spec 3.0 changes affect the bindings. Or if they even affect them at all.

Preliminary possible binding changes:
1. Operator changes, might affect HTTP bindings
2. Request reply, affect all bindings that has relevant properties for this.

This issue are meant as a placeholder until the changes in the spec are finalized.

jonaslagoni:
As the spec changes are nearly there, it's time to inspect the bindings. Tagging code owners.

@GeraldLoeffler do you have the possibility to inspect `anypointmq`?
@rcoppen do you have the bandwidth to inspect `ibmmq`?
@lbroudoux @dalelane do you folks have the bandwidth to inspect `kafka`?
@whitlockjc do you have the bandwidth to inspect`googlepubsub`?
@damaru-inc @CameronRushton do you folks have the bandwidth to inspect`solace`?
@VisualBean do you have the bandwidth to inspect`pulsar`?

jonaslagoni:
Here is the inspection of the other bindings:

amqp
- Operation binding object
  - `mandatory`, is message specific, but part of operation, not sure if this should change?
  - `replyTo`, with the request/reply change, this could probably be removed? BTW, send what response?

amqp1
- None

anypointmq
- Server Object
  - With the change to URL, the constraints change as well.
- Message Binding Object
  - Linking to schema object for headers, which should now be the custom header type.

jonaslagoni:
I am gonna create individual PRs for each protocol above for it to make sense on a review point of view.

whitlockjc:
I will gladly take care of the `googlepubsub` bindings. I'm not sure what, if anything, needs to be done right now.

jonaslagoni:
> I will gladly take care of the googlepubsub bindings. I'm not sure what, if anything, needs to be done right now.

Just had a quick look:
- Examples needs to change
- Message Binding Object - `schema` might change based on https://github.com/asyncapi/spec/pull/910 

But that's about it I think 🤔

jonaslagoni:
I created 6 PRs based on https://github.com/asyncapi/bindings/issues/182#issuecomment-1589239342

- https://github.com/asyncapi/bindings/pull/202
- https://github.com/asyncapi/bindings/pull/203
- https://github.com/asyncapi/bindings/pull/204
- https://github.com/asyncapi/bindings/pull/205
[...]

PR #207, opened 2023-06-13
feat: adapt websockets bindings to v3

jonaslagoni:
**Description**
This PR adapts the websockets bindings to v3. This first round of review is just for the general structure changes.

TODO right before final review:

- [x] Adapt examples, if any that contain v2 structure.
- [x] Figure out which version change this constitutes

TODO after merge:
- [ ] Adapt schema files in spec-schema-repo

**Related issue(s)**
Related to https://github.com/asyncapi/bindings/issues/182

### DIFF
--- a/websockets/README.md
+++ b/websockets/README.md
@@ -6,7 +6,7 @@ This document defines how to describe WebSockets-specific information on AsyncAP
 
 ## Version
 
-Current version is `0.1.0`.
+Current version is `1.0.0`.
 
 
 <a name="server"></a>
@@ -49,4 +49,4 @@ This object MUST NOT contain any properties. Its name is reserved for future use
 This object MUST NOT contain any properties. Its name is reserved for future use.
 
 
-[schemaObject]: https://www.asyncapi.com/docs/specifications/2.0.0/#schemaObject
+[schemaObject]: https://www.asyncapi.com/docs/specificationsv3.0.0-next-major-spec.12/#schemaObject
--- a/websockets/README.md
+++ b/websockets/README.md
@@ -6,17 +6,13 @@ This document defines how to describe WebSockets-specific information on AsyncAP
 
 ## Version
 
-Current version is `1.0.0`.
-
+Current version is `0.2.0`.
 
 <a name="server"></a>
 
 ## Server Binding Object
 
 This object MUST NOT contain any properties. Its name is reserved for future use.
-
-
-
 
 <a name="channel"></a>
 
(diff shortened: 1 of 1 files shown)


websockets/README.md
```diff
@@ -29,8 +25,8 @@ When using WebSockets, the channel represents the connection. Unlike other proto
 Field Name | Type | Description
 ---|:---:|---
 <a name="operationBindingObjectMethod"></a>`method` | string | The HTTP method to use when establishing the connection. Its value MUST be either `GET` or `POST`.
-<a name="operationBindingObjectQuery"></a>`query` | [Schema Object][schemaObject] | A Schema object containing the definitions for each query parameter. This schema MUST be of type `object` and have a `properties` key.
-<a name="operationBindingObjectHeaders"></a>`headers` | [Schema Object][schemaObject] | A Schema object containing the definitions of the HTTP headers to use when establishing the connection. This schema MUST be of type `object` and have a `properties` key.
+<a name="operationBindingObjectQuery"></a>`query` | [Parameters Object][parametersObject] | A Schema object containing the definitions for each query parameter. 
+<a name="operationBindingObjectHeaders"></a>`headers` | [Parameters Object][parametersObject] | A Schema object containing the definitions of the HTTP headers to use when establishing the connection.
```
fmvilas, line 29:
```suggestion
<a name="operationBindingObjectQuery"></a>`query` | [Parameters Object][parametersObject] | A Parameters object containing the definitions for each query parameter. 
<a name="operationBindingObjectHeaders"></a>`headers` | [Parameters Object][parametersObject] | A Parameters object containing the definitions of the HTTP headers to use when establishing the connection.
```

fmvilas (changes requested):
Still concerned about the usage of parameters for headers but left other suggestions too.

derberg (changes requested):
there is no need to release new version of the binding. The fact that we changed schema for parameters, do not mean we should change it for query params and headers. Even in the main spec we still allow schema object for headers.

`Parameters` from spec are simplified as they are focused on the content of AsyncAPI document. They are used to parametrise and describe dynamic parts of channel name. In case of query params, this is something that describes run time behaviour of the server, so it makes sense to for example specify `type: boolean` for the sake of better code generation -> https://github.com/asyncapi/spec/blob/master/examples/websocket-gemini.yml#L63-L96

jonaslagoni:
Alright, that makes sense 👍

PR #211, opened 2023-07-03
Create conventions

dpwdec:
**Description**

- ...
- ...
- ...

**Related issue(s)**


conventions.md
```diff
@@ -0,0 +1,301 @@
...
+
+A Binding may also be used by the providers of SDKs for MoM to provide metadata to configure producers or consumers using that SDK. That is outside the scope of advice here, but SDK owners wishing to use Bindings may follow the general advice here.
+
+## Effective Bindings
+### **Item 1** Use the Extensions Format
+Prefer use of the [Bindings Extensions](https://github.com/asyncapi/extensions-catalog) format when defining a binding over raw JSON Schema.
+
+The Binding Extensions format works with tools because it indicates where in the specification the binding should be used via a Hook. Using a Hook signals clearly to the user of an extension on which of Channel, Operation or Message the binding can be used. Implementers may choose to use author a binding so that the only fields on the binding are those appropriate to the object identified by the Hook. It also makes it possible for tools to determine correctness of an AsyncAPI file by ensuring that the binding is only used where permitted.
```
fmvilas, line 17: I don't think the "bindings extension format" will ever exist. This was a proof of concept that I did before we had bindings themselves. We should probably not recommend this here.

conventions.md
```diff
@@ -0,0 +1,301 @@
...
+
+### **Item** 2 Prefer JSON for Bindings
```
fmvilas, line 19:
```suggestion
### **Item 2** Prefer JSON for Bindings
```

conventions.md
```diff
@@ -0,0 +1,301 @@
...
+
+There may be multiple SDKs for a particular protocol, for different languages and frameworks. In many cases the variables required by these targets remain common across the targets and can simply be inferred from the protocol. In this case it is possible to define a common set of fields for a binding that can be used by Scaffolding regardless of the SDK. In other cases, SDKs vary considerably for the middleware. In this case you may prefer to provide a binding for the middleware that facilitates code generation for that middleware. 
+
+### **Item 5** Assume Separate Endpoints Have Separate AsyncAPI Files
+Producers and consumers are usually separate apps, they may belong to separate teams, and under models such as pub-sub, there may an architectural goal to decouple producer and consumer by ensuring that each knows about the channel, but not the other. For this reason, you should assume that it will be common to define separate endpoints for the producer and consumer, but that the channel and the message schema will be shared. (Users may use $ref to prevent duplication, but that does not change these guidelines).
```
fmvilas, line 45:
```suggestion
Producers and consumers are usually separate apps, they may belong to separate teams, and under models such as pub-sub, there may be an architectural goal to decouple producer and consumer by ensuring that each knows about the channel, but not the other. For this reason, you should assume that it will be common to define separate endpoints for the producer and consumer, but that the channel and the message schema will be shared. (Users may use $ref to prevent duplication, but that does not change these guidelines).
```

conventions.md
```diff
@@ -0,0 +1,301 @@
...
+
+Where a protocol defines metadata that forms part of the message sent via middleware it should be defined as part of the header schema (potentially as MessageTraits), and where the protocol defines data (or transmits metadata in the payload) it should be defined as part of the payload. Note that the payload does not support MessageTraits. 
+
+Tooling to validate that messages match the schema will expect to verify against the schema of the message payload and headers, putting message metadata into a Message Binding makes this harder as tooling must know to look there as well. 
+
+Implementers of producers or consumers are also likely to assume that the the can refer to #/Components/Schema to understand payloads and headers, and adding metadata into bindings in addition makes it harder to comprehend the schema of a message.
```
fmvilas, line 198: This is not true. The right place to set protocol-specific message properties is actually the bindings, not traits or headers. See the definition of the [Message headers field](https://www.asyncapi.com/docs/reference/specification/v2.6.0#messageObject):

> Schema definition of the application headers. Schema MUST be of type "object". **It MUST NOT define the protocol headers**.

The reason this is separated is to prevent mixing protocol-specific and application-specific metadata in a single place and the user having to define the shape of the metadata with JSON Schema every time.

My suggestion is that we remove the whole Item 9.

conventions.md
```diff
@@ -0,0 +1,301 @@
...
+      payload:
+        $ref: "#/components/schemas/quux"
+      traits:
+        $ref: "#/components/schemas/amqpBasicProperties" 
+
+
+```
+
+### **Item 10** Be Consistent with Other Protocols
+Design your binding with the possibility that a producer might expose a message over different bindings - so as to support a wider range of consumers. By following the guidelines in this document your binding should be compatible with other bindings, and simplify the development of tooling that supports those because they operate at the same level.
```
fmvilas, line 263: > your binding should be compatible with other bindings

How exactly can someone accomplish this? We don't have a common structure among bindings yet.

fmvilas (changes requested):
This is a great initiative! Left some suggestions.

smoya:
@dpwdec Is this PR supposed to be a replacement for https://github.com/asyncapi/bindings/pull/211 ?

smoya:
ping @dpwdec :D

PR #214, opened 2023-09-04
docs: updated sqs binding readme file

gokerakc:
**Description**

- `type: standard` is not a valid property. Based on the SQS schema we need to use the property **fifoQueue**.

### DIFF
--- a/sqs/README.md
+++ b/sqs/README.md
@@ -127,7 +127,7 @@ channels:
       sqs:
         queue:
           name: user-signedup-queue
-          type: standard
+          fifoQueue: false
           receiveMessageWaitTime: 4
           redrivePolicy:
             deadLetterQueue:
@@ -143,7 +143,7 @@ channels:
         deadLetterQueue: 
           name: user-signedup-dlq 
           messageRetentionPeriod: 1209600 
-	        type: standard
+	        fifoQueue: false
     subscribe:
       operationId: sendMessage
       description: sends messages when a user has signed up
@@ -218,7 +218,7 @@ channels:
         sqs:
           queues:
           - name: user-signedup-queue
-            type: standard
+            fifoQueue: false
             receiveMessageWaitTime: 4
             policy:
               statements: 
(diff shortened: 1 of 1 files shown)


iancooper (changes requested):
AWS supports two queue types for SQS: standard and FIFO. Please see: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-queue-types.html

Is there a problem underlying this request?

gokerakc:
I've edited the PR description to explain the issue better.

dpwdec:
@iancooper we changed the schema to [reflect the AWS API](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sqs-queue.html#aws-resource-sqs-queue-properties) for making a queue FIFO or Standard which is done via a boolean called FifoQueue. In the AsyncAPI schema [here](https://github.com/asyncapi/bindings/blob/a7f7f64587377d9fd45cbc4c8d09de96c7280dab/sqs/json_schemas/channel.json#L45)

I believe @gokerakc ‘s PR is just bringing the examples in line with the schema because these were missed on the initial commit for the SQS binding and still show the original design.

iancooper:
Thanks for that @gokerakc and @dpwdec. Type only really has advantages if AWS issues another type, so that we cannot use a boolean flag. That is however speculative, and given that AWS would have to adjust their API to fix the issue, I think we can agree to opt for what would be most obvious for an AWS developer, in this case using the boolean flag instead of a type field.

Thanks for the explanation.

iancooper (approved):

dpwdec (approved):

derberg:
@iancooper @dpwdec I guess this one is ok to be merged?

derberg:
@dpwdec fyi, next time you can also merge with `/rtm` comment under a PR. This way you are sure that merging is done 100% accurate with proper PR. Most important, less clicking 😄

PR #215, opened 2023-09-18
docs: update FilterPolicy config to match AWS API

Gadam8:
Currently, filterPolicy configuration outlined in this file does not match what is outlined in the AWS documentation and API - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sns-subscription.html. The `filterPolicy` should only contain the JSON policy, so no need for the `attributes` property; which in itself was confusing as the policy can apply to the message attributes or body.

Further to this, this definition file was missing the `filterPolicyScope` which is needed to determine whether the user wishes to filter on the message attributes or the body.

**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/sns/json_schemas/operation.json
+++ b/sns/json_schemas/operation.json
@@ -96,7 +96,7 @@
         },
         "filterPolicy": {
           "type": "object",
-          "description": "Only receive a subset of messages from the channel, determined by this policy. Depending on the FilterPolicyScope, a map of either a message attribute or message body to an array of possible matches. The match may be a simple string for an exact match, but it may also be an object that represents a constraint and values for that constraint.",
+          "description": "Only receive a subset of messages from the channel, determined by this policy. Depending on the filterPolicyScope, a map of either a message attribute or message body to an array of possible matches. The match may be a simple string for an exact match, but it may also be an object that represents a constraint and values for that constraint.",
           "patternProperties": {
             "^x-[\\w\\d\\.\\-\\_]+$": {
               "$ref": "https://raw.githubusercontent.com/asyncapi/spec-json-schemas/v2.14.0/schemas/2.4.0.json#/definitions/specificationExtension"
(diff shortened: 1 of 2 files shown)


sns/json_schemas/operation.json
```diff
@@ -97,6 +97,9 @@
         "filterPolicy": {
           "$ref": "#/definitions/filterPolicy"
         },
+        "filterPolicyScope": {
```
dpwdec, line 100: As `filterPolicy` and `filterPolicyScope` will be properties *directly* on the object they don't need to be `$ref`'d.

dpwdec (approved):

iancooper (dismissed):

iancooper:
@dpwdec Approved

VisualBean:
>  whether the user wishes to filter on the message attributes or the body.

sounds odd to me.
The bindings is not for users to create, it for users to consume and learn more about the topic. I dont think consumer configuration should be a part of it.

Gadam8:
> > whether the user wishes to filter on the message attributes or the body.
> 
> sounds odd to me. The bindings is not for users to create, it for users to consume and learn more about the topic. I dont think consumer configuration should be a part of it.
> 
> Are filterpolicies set by the event producer to be used by the consumer?

Although the SNS subscription lives under the SNS remit it is under the control of the consumer of the topic. Filter policies are set by the consumer of the event, not the producer. It is the consumer who is deciding how they wish to consume the events sent by the producer, not the producer determining how their events should be consumed.

VisualBean:
Okay, 
Does it make sense to communicate to consumers of the  specification then? 
I don't immediately see the value in communicating "we filter events based on this" - but maybe I'm missing something

dpwdec (approved):

Gadam8:
Thanks @dpwdec and @iancooper for the approval. @KhudaDad414 can I get a review please?

KhudaDad414 (approved):
Schemas seems to match with README.

derberg:
@dpwdec @KhudaDad414 just n info for future -  that when you change structure of binding, you should update binding version too

PR #228, opened 2023-11-30
feat: enable additional properties for topicConfigurationObject

gokerakc:
**Description**

- Updated **additionalProperties** as **true** for the **topicConfigurationObject** due to the existence of several vendor-specific configuration properties (e.g. confluent.schema.validation, confluent.value.subject.name.strategy, confluent.key.subject.name.strategy).
- This change was proposed in one of the old [issues](https://github.com/asyncapi/bindings/issues/163) and it was approved but I believe the **additionalProperties** bit got missed by mistake.
- Please let me know if there is any obstacle to enabling that.

[](https://github.com/asyncapi/bindings/issues/163)

### DIFF
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -6,7 +6,7 @@ This document defines how to describe Kafka-specific information on AsyncAPI.
 
 ## Version
 
-Current version is `0.4.0`.
+Current version is `0.5.0`.
 
 
 <a name="server"></a>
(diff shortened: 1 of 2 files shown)


lbroudoux:
Hey there!

I understand the general purpose of enabling additional properties, but - for this specific concern of schema and registry-related things, we already have `schemaRegistryVendor`, `schemaLookupStrategy ` and others `schema*`  properties.
Shouldn't we try to encourage using those (maybe adding some missing stuff) in a more vendor-neutral and generic way? 

BTW: we'll also have to report this `additionalProperties` activation to the v3 json-schemas as well.

KhudaDad414 (approved):
LGTM!

derberg:
@dalelane @lbroudoux folks, today evening we merge `next-major-spec` to `master` so if you want to merge this one before, this is the right time.

derberg:
and if it comes to versioning, yes, imho it is new version because you also need to adjust `kafka` binding readme -> https://github.com/asyncapi/bindings/tree/master/kafka#topicconfiguration-object as now it says that only "listed" properties are allowed, and now you change that it can be extended with more properties.

anyway, not much time left 😉

gokerakc:
Thanks @lbroudoux!

IMO, the new property `schemaLookupStrategyForKey` sounds good but if we need to be quick because of the `next-major-spec` branch merging work I would say let's add the new properties in another PR so we can have more time for thinking about naming and also possible other properties.

derberg:
Hey folks, looks like there is not 💯 consensus, so we had a chat and do not want to rush here and merge this one as it would complicate life a bit -> https://asyncapi.slack.com/archives/C0230UAM6R3/p1701202455849719 (to join slack go to https://asyncapi.com/slack-invite)

Please continue discussion after we release v3 and improve what you need. From perspective of AsyncAPI v2 bindings were "experimental" and we did not validate them - AsyncAPI 3.0.0 is the first release where specification JSON Schema file includes bindings JSON Schema. So tooling do not block you from using additional properties for kafka binding in AsyncAPI v2 documents

PR #233, opened 2024-01-08
docs: fix incorrect example for AMQP

jonaslagoni:
According to the JSON Schema files, you are only allowed to use exchange or queue depending on the `is`: https://github.com/asyncapi/spec-json-schemas/blob/287ed2846a764ac6ffce1d5034a9e3e47853a247/bindings/amqp/0.3.0/channel.json#L84

You cannot have both, which makes the example invalid.

### DIFF
--- a/amqp/README.md
+++ b/amqp/README.md
@@ -54,12 +54,6 @@ channels:
     bindings:
       amqp:
         is: routingKey
-        queue:
-          name: my-queue-name
-          durable: true
-          exclusive: true
-          autoDelete: false
-          vhost: /
         exchange:
           name: myExchange
           type: topic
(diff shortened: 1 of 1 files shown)


Tenischev:
@jonaslagoni actually, having both makes sense and I would say that the description is wrong, not the example

jonaslagoni:
@Tenischev might be, I have no idea 😄 I would definitely do a new issue discussing this. 

This PR just fixes the example based on the current version of AMQP 🙂

g-radam:
Is someone able to elaborate how you would declare an AMQP channel queue binding, where that queue is bound to an exchange (in the RabbitMQ sense)? See: [RabbitMQ Queue Binding](https://www.rabbitmq.com/tutorials/tutorial-three-javascript#bindings).

I interpreted the original example:
```
channels:
  userSignup:
    address: 'user/signup'
    bindings:
      amqp:
        is: queue
        queue:
          name: my-queue-name
          durable: true
          exclusive: true
          autoDelete: false
          vhost: /
        exchange:
          name: my-service-exchange
        bindingVersion: 0.3.0
```

to mean the channel declares a queue called "user/signup" which operations will use, and that queue was **bound** to an exchange of my choice (IE receive messages from that exchange). Unless my interpretation of this is wrong, how else would you declare that the channel queue is bound to a custom exchange??

I would have said the amqp schema is wrong, not the example. Am I missing something here?

Issue #234, opened 2024-01-10
How to define response code for HTTP


jonaslagoni:
#### Reason/Context
Right now there are no way to define response codes for a message/operation for HTTP. 

Right now I am doing it like this for messages: 

```
...
      HttpCustomersGetMessageResponse:
        description: OK
        payload:
          $ref: '#/components/schemas/CustomerPaginated'
        bindings:
          http:
            x-response-code: 200
```

However, I think this should be part of the official bindings in some way or form

PR #235, opened 2024-01-11
feat: add HTTP status code

jonaslagoni:
**Description**
This PR adds the `statusCode` property to the message object.

The reason for adding it to the message object and not the operation object, is that it feels more natural, as the status code belong to each message being send.
```
...
channels:
  HttpCustomersChannel:
    address: /customers
    messages:
      HttpCustomersGetMessageResponse:
        description: OK
        payload:
          $ref: '#/components/schemas/CustomerPaginated'
        bindings:
          http:
            statusCode: 200
            bindingVersion: '0.3.0'

      HttpCustomersGetMessageResponse:
        description: Server error, could not complete request
        payload:
          $ref: '#/components/schemas/SomeError'
        bindings:
          http:
            statusCode: 500
            bindingVersion: '0.3.0'

operations:
  GetCustomers:
    action: send
    channel:
       $ref: #/channels/HttpCustomersChannel
    bindings:
      http:
        method: GET
        bindingVersion: '0.3.0'
```

**Related issue(s)**
Fixes https://github.com/asyncapi/bindings/issues/234

### DIFF
--- a/http/README.md
+++ b/http/README.md
@@ -71,7 +71,7 @@ This object contains information about the message representation in HTTP.
 Field Name | Type | Description
 ---|:---:|---
 <a name="messageBindingObjectHeaders"></a>`headers` | [Schema Object][schemaObject] \| [Reference Object](referenceObject) | A Schema object containing the definitions for HTTP-specific headers. This schema MUST be of type `object` and have a `properties` key.
-<a name="messageBindingObjectStatusCode"></a>`statusCode` | number | The HTTP response status code according to [RFC 9110](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes).
+<a name="messageBindingObjectStatusCode"></a>`statusCode` | number | The HTTP response status code according to [RFC 9110](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes). `statusCode` is only relevant for messages referenced by the [operation reply object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationReplyObject) as they define the status code for the response, in all other cases this can safely be ignored. 
 <a name="messageBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
 
 This object MUST contain only the properties defined above.
--- a/http/README.md
+++ b/http/README.md
@@ -21,7 +21,6 @@ This object MUST NOT contain any properties. Its name is reserved for future use
 
 This object MUST NOT contain any properties. Its name is reserved for future use.
 
-
 <a name="operation"></a>
 
 ## Operation Binding Object
@@ -61,7 +60,6 @@ operations:
         bindingVersion: '0.3.0'
 ```
 
-
 <a name="message"></a>
 
 ## Message Binding Object
@@ -73,11 +71,10 @@ This object contains information about the message representation in HTTP.
 Field Name | Type | Description
 ---|:---:|---
 <a name="messageBindingObjectHeaders"></a>`headers` | [Schema Object][schemaObject] \| [Reference Object](referenceObject) | A Schema object containing the definitions for HTTP-specific headers. This schema MUST be of type `object` and have a `properties` key.
-<a name="messageBindingObjectStatusCode"></a>`status-code` | number | The HTTP response status code according to [RFC 9110](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes).
+<a name="messageBindingObjectStatusCode"></a>`statusCode` | number | The HTTP response status code according to [RFC 9110](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes).
 <a name="messageBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
 
 This object MUST contain only the properties defined above.
-
 
 ```yaml
 channels:
@@ -87,7 +84,7 @@ channels:
       testMessage:
         bindings:
           http:
-            status-code: 200
+            statusCode: 200
             headers:
               type: object
               properties:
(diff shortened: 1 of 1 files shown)


smoya:
I understand the status code can't exist in the operation since in one channel can coexist several messages (responses) with a different status code each (200, 202, 500...). You won't be able to do this within a single operation.

smoya (dismissed):
LGTM! 🚀🌔

smoya:
Wondering how this will work with Operation Reply objects 🤔

http/README.md
```diff
@@ -73,6 +73,7 @@ This object contains information about the message representation in HTTP.
 Field Name | Type | Description
 ---|:---:|---
 <a name="messageBindingObjectHeaders"></a>`headers` | [Schema Object][schemaObject] \| [Reference Object](referenceObject) | A Schema object containing the definitions for HTTP-specific headers. This schema MUST be of type `object` and have a `properties` key.
+<a name="messageBindingObjectStatusCode"></a>`status-code` | number | The HTTP response status code according to [RFC 9110](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes).
```
fmvilas, line 76: We're using camel case everywhere else:

```suggestion
<a name="messageBindingObjectStatusCode"></a>`statusCode` | number | The HTTP response status code according to [RFC 9110](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes).
```

fmvilas (changes requested):

http/README.md
jonaslagoni: Applied :+1:

jonaslagoni:
When you define your application as a REST server i.e. you receive a request where you respond with a reply that has specific status codes it looks like this:

```yaml
asyncapi: 3.0.0
info:
  title: REST server
  version: 0.0.1
defaultContentType: application/json
servers:
  HttpServer:
    host: 'http://localhost:8080/api'
    protocol: http
    security:
      - type: http
        scheme: basic
      - type: http
        scheme: bearer
        bearerFormat: JWT
channels:
  HttpCustomersChannel:
    address: /customers
    servers:
      - $ref: '#/servers/HttpServer'
    messages:
      HttpCustomersGetMessage:
        payload:
          type: 'null'
      HttpCustomersGetMessageResponse:
        description: OK
        payload:
          $ref: '#/components/schemas/CustomerPaginated'
        bindings:
          http:
            statusCode: 200
operations:
  onListCustomers:
    action: receive
    description: listCustomers
    channel:
      $ref: '#/channels/HttpCustomersChannel'
    messages:
      - $ref: '#/channels/HttpCustomersChannel/messages/HttpCustomersGetMessage'
    reply:
      channel:
        $ref: '#/channels/HttpCustomersChannel'
      messages:
        - $ref: >-
            #/channels/HttpCustomersChannel/messages/HttpCustomersGetMessageResponse
components:
  schemas:
    Customer:
      type: object
    CustomerPaginated:
      type: object
```

When you define your application as interacting with a REST server as a client where are the one sending the request, nothing changes for the messages nor the status code 🤔 
```yaml
asyncapi: 3.0.0
info:
  title: REST client
  version: 0.0.1
defaultContentType: application/json
servers:
  HttpServer:
    host: 'http://localhost:8080/api'
    protocol: http
    security:
      - type: http
        scheme: basic
      - type: http
        scheme: bearer
        bearerFormat: JWT
channels:
  HttpCustomersChannel:
    address: /customers
    servers:
      - $ref: '#/servers/HttpServer'
    messages:
      HttpCustomersGetMessage:
        payload:
          type: 'null'
      HttpCustomersGetMessageResponse:
        description: OK
        payload:
          $ref: '#/components/schemas/CustomerPaginated'
        bindings:
          http:
            statusCode: 200
operations:
  onListCustomers:
    action: send
    description: listCustomers
    channel:
      $ref: '#/channels/HttpCustomersChannel'
    messages:
      - $ref: '#/channels/HttpCustomersChannel/messages/HttpCustomersGetMessage'
    reply:
      channel:
        $ref: '#/channels/HttpCustomersChannel'
      messages:
        - $ref: >-
            #/channels/HttpCustomersChannel/messages/HttpCustomersGetMessageResponse
components:
  schemas:
    Customer:
      type: object
    CustomerPaginated:
      type: object
```

In short, it just means that messages referenced in the operation should NOT have statusCode defined, but messages defined in the reply section SHOULD 🤔 

How do you folks suggest we clarify the interaction here in a meaningful way 🤔 ?

fmvilas:
> In short, it just means that messages referenced in the operation should NOT have statusCode defined, but messages defined in the reply section SHOULD 🤔

I think we can say that statusCode in the operation messages (not reply) can simply be ignored. We only take it into account inside the `reply` section. How does that sound?

jonaslagoni:
@fmvilas added the following as description (feel free to reformat and change it as needed):

```
The HTTP response status code according to [RFC 9110](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes). `statusCode` is only relevant for messages referenced by the [operation reply object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationReplyObject) as they define the status code for the response, in all other cases this can safely be ignored.
```

fmvilas (approved):

jonaslagoni:
@smoya anything from your side?

smoya (approved):
> > In short, it just means that messages referenced in the operation should NOT have statusCode defined, but messages defined in the reply section SHOULD 🤔
> 
> I think we can say that statusCode in the operation messages (not reply) can simply be ignored. We only take it into account inside the `reply` section. How does that sound?

I don't feel this is very user friendly but I can't find a better alternative considering the status code SHOULD be located in the Message Object because the reasons already mentioned.

So 👍 in my side.

jonaslagoni:
Added the JSON schema changes here: https://github.com/asyncapi/spec-json-schemas/pull/484

Feel free to merge in any order you see fit 🙂

PR #237, opened 2024-01-27
feat: solace add per operation fields

GreenRover:
**Description**

- add new field:  timeToLive on a per operation bases.  
  To enable to documentate what the a good valid related to the business logic of this event. For example if a temperature reading is publish every 30 sec. A ttl >30sec would make no sense.  
  I choose  a per operation bases to align to https://github.com/asyncapi/bindings/blob/master/mqtt/README.md
- add new field: priority  on a per operation bases.  
  To enable to documentate the importent ratings of events.
- add new field: dmqEligible on a per operation bases.  
  To enable a message consumer to provide all necessary Information how the expected messages should look like.

### DIFF
--- a/solace/README.md
+++ b/solace/README.md
@@ -6,7 +6,7 @@ This document defines how to describe Solace-specific information with AsyncAPI.
 
 ## Version
 
-Current version is `0.3.0`.
+Current version is `0.4.0`.
 
 <a name="server"></a>
 
@@ -14,8 +14,9 @@ Current version is `0.3.0`.
 
 Field Name | Type | Description
 ---|---|---
-`bindingVersion`|String|The current version is 0.3.0
+`bindingVersion`|String|The current version is 0.4.0
 `msgVpn`|String|The Virtual Private Network name on the Solace broker.
+`clientName`|String|A unique client name to use to register to the appliance. If specified, it must be a valid Topic name, and a maximum of 160 bytes in length when encoded as UTF-8.
 
 
 <a name="channel"></a>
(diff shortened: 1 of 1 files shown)


solace/README.md
```diff
@@ -32,10 +33,13 @@ This object MUST NOT contain any properties. Its name is reserved for future use
 
 We need the ability to support several bindings for each operation, see the [Example](#example) section below for details.
 
-Field Name | Type | Description
----|---|---
-`bindingVersion`|String|The current version is 0.3.0
-`destinations`|List of Destination Objects|Destination Objects are described next.
+Field Name | Type                                                                            | Description
+---|---------------------------------------------------------------------------------|---
+`bindingVersion`| String                                                                          |The current version is 0.3.0
```
CameronRushton, line 38: Please update the binding version to 0.4.0 throughout the document.

solace/README.md
```diff
@@ -32,10 +33,13 @@ This object MUST NOT contain any properties. Its name is reserved for future use
...
+`destinations`| List of Destination Objects                                                     |Destination Objects are described next.
+`timeToLive`  | Integer \| [Schema Object][schemaObject] \| [Reference Object](referenceObject) | Interval in milliseconds or a *Schema Object* containing the definition of the lifetime of the message. 
+`priority`  | Integer \| [Schema Object][schemaObject] \| [Reference Object](referenceObject) | The valid priority value range is 0-255 with 0 as the lowest priority and 255 as the highest or a *Schema Object* containing the definition of the priority.
```
CameronRushton, line 41: >[Schema Object][schemaObject] \| [Reference Object](referenceObject)

This appears to be broken. I think for v3, it's supposed to go here for referenceObject (https://www.asyncapi.com/docs/reference/specification/v3.0.0#referenceObject). Similarly for schemaObject.

CameronRushton (changes requested):
Thanks for this. A couple simple requests to address & thank you for volunteering to update spec-json-schemas as well.

solace/README.md
GreenRover: changed

solace/README.md
GreenRover: thanks for pointing out. i missed to links. Added it at the bottm.

GreenRover:
I changed example to async api 3.0 in same pr to avoid a version incretion to 0.5.0 for that

CameronRushton (approved):
Looks great. We really appreciate your help!

Issue #231, opened 2023-12-12
Enable additionalProperties for kafka topicConfigurationObject


gokerakc:
#### Reason/Context

Currently, only 5 topic configuration properties are supported by the asyncapi kafka binding (json schema [here](https://github.com/asyncapi/spec-json-schemas/blob/master/bindings/kafka/0.4.0/channel.json#L28)).

These properties are;

- **cleanup.policy**
- **retention.ms**
- **retention.bytes**
- **delete.retention.ms**
- **max.message.bytes**

But in some cases, there is a need to specify vendor-specific topic configurations (e.g. [confluent-topic-configs](https://docs.confluent.io/platform/current/installation/configuration/topic-configs.html)). For that reason I think it makes sense to increase the version of that binding and enable **additionalProperties** for the **topicConfigurationObject**.

In my use case, I need to use the following additional properties and they can also be added to the **topicConfigurationObject** if you see a good fit. Please see the properties from the vendor and proposed property names for the **topicConfigurationObject**:

- confluent.key.schema.validation (`schema.key.validation.enabled`)
- confluent.key.subject.name.strategy (`schema.key.lookup.strategy`)
- confluent.value.schema.validation (`schema.value.validation.enabled`)
- confluent.value.subject.name.strategy (`schema.value.lookup.strategy`)

Please share your thoughts about enabling additional properties and adding these optional properties to the object.

gokerakc:
@lbroudoux I would like to get your thoughts on this because you [commented](https://github.com/asyncapi/bindings/pull/228#issuecomment-1839488931) about adding new properties before.

derberg:
looks like completed

PR #238, opened 2024-01-29
feat: updated kafka topic configuration object

gokerakc:
**Description**

- Updated additionalProperties flag as true
- Added 4 new properties to the topic configuration object
- Please see the related [issue](https://github.com/asyncapi/bindings/issues/231) for more details.

**Related issue(s)**
https://github.com/asyncapi/bindings/issues/231

### DIFF
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -6,7 +6,7 @@ This document defines how to describe Kafka-specific information on AsyncAPI.
 
 ## Version
 
-Current version is `0.4.0`.
+Current version is `0.5.0`.
 
 
 <a name="server"></a>
@@ -32,7 +32,7 @@ servers:
       kafka:
         schemaRegistryUrl: 'https://my-schema-registry.com'
         schemaRegistryVendor: 'confluent'
-        bindingVersion: '0.4.0'
+        bindingVersion: '0.5.0'
 ```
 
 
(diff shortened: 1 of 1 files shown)


dalelane (commented):
I would rather not bring vendor-specific extensions into the bindings like this. 

Confluent are pretty good at keeping their platform-specific extensions, that aren't part of Kafka, separate - by prefixing the option names with `confluent.` 

For example, they have `confluent.key.schema.validation` that you have described here as `schema.key.validation.enabled`. 

I think that is a good pattern to adopt - while I absolutely see value in us including widely used vendor extensions in the bindings, I think they should be clearly identified as such. Not only that, if we match the names to what they are to use, we make the life of developers easier as they don't need to translate between naming formats. 

I recommend copying the property names as-is. If you think the renaming is helpful (e.g. adding "enabled" to make it clearer it is a boolean) then I can live with that but I'd still prefer that we keep the "confluent." prefix.

gokerakc:
Hi @dalelane,

I've updated the property names as you suggested. It makes sense to keep that widely used vendor-specific config name as it is since it'll help developers when it comes to mapping those values.

In general, I'm in favour of not leaking vendor-specific properties into the asyncapi bindings but I think it's acceptable to add some of the widely used ones and update **additionalProperties** as **true** for the less popular configs so toolings can consider parsing additional properties for the topicConfiguration object.

dalelane (approved):
looks good, thanks for the updates

gokerakc:
Thanks for approving the PR. Once this one is merged I'll open the json schema PR (https://github.com/asyncapi/spec-json-schemas/pull/481) for review. @dalelane

gokerakc:
Hi @derberg, I think this PR is ready for merge. Do you want to merge it or wait for more reviewers?

gokerakc:
Hey @lbroudoux, would you like to review/merge these changes?

PR #242, opened 2024-02-13
fix: updated topic configuration object data types

gokerakc:
**Description**

The following properties need to be updated as `long`:
- **retention.ms**

- **retention.bytes**

- **delete.retention.ms**

If you need further details about the configs you can check this documentation page: https://kafka.apache.org/documentation/#topicconfigs_retention.bytes

**Change log**
- Updated data types from **integer** to **long**


dalelane (commented):
I'm absolutely not a schema expert here, but I thought we're speaking the language of JSON schema here, which means we're restricted to "integer" as a way of describing whole numbers. 

e.g. https://json-schema.org/understanding-json-schema/reference/numeric 

Happy for you to explain why I'm wrong though :)

gokerakc:
If we are dependent on JSON in this README.md file then you are right, we can not use **long** as a data type.

But also it would be good to raise that developers should use **long** in their tooling otherwise they may end up having problems.

dalelane:
That's a fair point - I'm perhaps being unhelpfully strict for a markdown file, and this is better thought of as an issue for https://github.com/asyncapi/spec-json-schemas only, where you do have them as integers. 

@smoya - what do you think?

smoya:
> That's a fair point - I'm perhaps being unhelpfully strict for a markdown file, and this is better thought of as an issue for https://github.com/asyncapi/spec-json-schemas only, where you do have them as integers.
> 
> @smoya - what do you think?

The issue is that you can't specify `long` as a type in JSON Schema, so we won't be able to really ensure it. 
Anyway, I also agree this README should mention `long` to be more accurate.

dalelane (approved):

PR #252, opened 2024-06-07
feat: support `sse` protocol

jfallows:
**Description**
- Added `sse` message binding object
- Uses `http` protocol in server definition

Note: we are in the process of testing these proposed changes with `zilla` open source project to confirm they are suffcient to support `sse` message validation, etc.

See https://github.com/aklivity/zilla/issues/952

### DIFF
--- a/sse/README.md
+++ b/sse/README.md
@@ -51,9 +51,9 @@ This object contains information about the message representation in SSE.
 ##### Fixed Fields
 
 Field Name | Type | Description
----|:---:|:---:|---
-<a name="messageBindingObjectEventType"></a>`event` | string | Server-sent event type, defaults to `message` if omitted, by the [SSE specification](protocolSpecification). |
-<a name="messageBindingObjectBindingVersion"></a>`bindingVersion` | string | | The version of this binding. If omitted, "latest" MUST be assumed.
+---|:---:|---
+<a name="messageBindingObjectEventType"></a>`event` | string | Server-sent event type, defaults to `message` if omitted, by the [SSE specification][protocolSpecification]. |
+<a name="messageBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
 
 This object MUST contain only the properties defined above.
 
--- a/sse/README.md
+++ b/sse/README.md
@@ -4,7 +4,7 @@ This document defines how to describe Server Sent Events-specific information on
 
 See the [Server-Sent Events protocol specification][protocolSpecification].
 
-Server-Sent Events requires `http` protocol, so the SSE [Message Binding Object](#message) below is appropriate for Servers with protocol `http`.
+Server-Sent Events requires `http` protocol, so the SSE [Message Binding Object](#message) below is appropriate for AsyncAPI Servers with protocol `http` or `https`.
 
 <a name="version"></a>
 
--- a/sse/README.md
+++ b/sse/README.md
@@ -28,7 +28,6 @@ When using Server Sent Events, the channel represents a single logical message s
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="operationBindingObjectMethod"></a>`method` | string | The HTTP method to use when establishing the request. Its value MUST be `GET`.
 <a name="operationBindingObjectQuery"></a>`query` | [Schema Object][schemaObject] \| [Reference Object](referenceObject) | A Schema object containing the definitions for each query parameter. This schema MUST be of type `object` and have a `properties` key.
 <a name="operationBindingObjectHeaders"></a>`headers` | [Schema Object][schemaObject] \| [Reference Object](referenceObject) | A Schema object containing the definitions of the HTTP headers to use when establishing the request. This schema MUST be of type `object` and have a `properties` key.
 <a name="operationBindingObjectBindingVersion"></a>`bindingVersion` | string | The version of this binding. If omitted, "latest" MUST be assumed.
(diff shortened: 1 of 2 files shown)


fmvilas:
Shouldn't this fall under the HTTP binding instead?

sse/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Channel Binding Object
+
+When using Server Sent Events, the channel represents a single logical message stream flowing from server to client. 
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---:|---
+<a name="operationBindingObjectMethod"></a>`method` | string | The HTTP method to use when establishing the request. Its value MUST be `GET`.
```
fmvilas, line 31: If the only allowed value is `GET`, can we just remove this field?

On the other hand, this whole object looks almost identical to the HTTP one. Shall we use the HTTP binding instead?

sse/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Message Binding Object
+
+This object contains information about the message representation in SSE.
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---:|---
+<a name="messageBindingObjectEventType"></a>`event` | string | Server-sent event type, defaults to `message` if omitted, by the [SSE specification][protocolSpecification]. |
```
fmvilas, line 55: I think this is the only part of this binding that's truly related to SSE. The rest is HTTP. So my proposal is that you add this info to the HTTP message binding instead. We already have `headers` and `statusCode`. We can add `sse` which in turn will contain `eventType`.

fmvilas (changes requested):

jonaslagoni:
Would users reading the bindings, know that the SSE-specific information is part of the HTTP bindings? Or is it more for convenience in maintaining them? 🤔 

I mean leaning more towards keeping them separate I think 🤔

jfallows:
Hey guys, thanks for the feedback. 

SSE does have a standard content type for http responses, defining the SSE protocol framing, but the message payload defined by the application is contained in that framing, i.e. one or more data fields per SSE event. 

Therefore message schema types for validation need to be described at a higher level than the SSE http response, as SSE has own logical channel of application defined messages.

sse/README.md
jfallows: The optional `last-event-id` request header is implicitly defined for SSE requests.

Actually, this header represents a cursor indicating progress into the SSE stream, allowing a disconnected client to recover from where they left off. So the logical SSE message stream is potentially spanning multiple HTTP requests.

If not abstracted out by `sse` protocol, how would the definition of acceptable HTTP headers that be handled by the proposed approach to use the HTTP binding for SSE message streams?

jfallows: > If the only allowed value is GET, can we just remove this field?
Yes. 👍

sse/README.md
jfallows: Agreed this is majority of it, though there is protocol framing in the HTTP response, and the optional `last-event-id` request header is implicitly defined for SSE requests.

Actually, this header represents a cursor indicating progress into the SSE stream, allowing a disconnected client to recover from where they left off. So the logical SSE message stream is potentially spanning multiple HTTP requests.

If not abstracted out by `sse` protocol, how would the definition of acceptable HTTP headers that be handled by the proposed approach to use the HTTP binding for SSE message streams?

fmvilas: Fair enough. I've never used SSE myself so I'm just speaking from a theoretical point of view.

fmvilas:
> as SSE has its own logical channel of application defined messages

🤔 This wasn't obvious to me at first glance. Can't we map AsyncAPI channels to these logical channel definitions? Cause I have the feeling that's how it should be but speaking from pure gut feeling.

jfallows:
> > as SSE has its own logical channel of application defined messages
> 
> 🤔 This wasn't obvious to me at first glance. Can't we map AsyncAPI channels to these logical channel definitions? Cause I have the feeling that's how it should be but speaking from pure gut feeling.

Right, the endpoint is identified via URL such as `https://example.com/server/sent/events`.

Here we propose to identify the logical channel of messages via the request path, such as `/server/sent/events` above, much like the `http` and `websockets` bindings.

fmvilas:
Then I think http://example.com should be a server definition and `/server/sent/events` should be the address of the channel. So better to keep it out of the sse binding. What do you think?

jfallows:
> Then I think http://example.com should be a server definition and `/server/sent/events` should be the address of the channel. So better to keep it out of the sse binding. What do you think?

`https://example.com/server/sent/events` is the endpoint URL from the clients point of view, such as via JavaScript in the browser.
```
var events = new EventSource("https://example.com/server/sent/events");
```

Are you suggesting that the server definition would use `http` protocol instead of `sse` protocol?

If so, then we would have just the `Message Binding Object`, indicating the `sse` `event` type such as `message` to specify allowed `sse` message types?

fmvilas:
Yeah, exactly. That would simplify things IMHO. The server protocol could be `http` or `http+sse`. Either way should be fine.

jfallows:
> Yeah, exactly. That would simplify things IMHO. The server protocol could be `http` or `http+sse`. Either way should be fine.

@fmvilas  I've gone with `http` for now, since there are more dimensions to this with secure and `http` or `https` seems simpler.

Please let me know if having an `sse` message binding object without specifying the protocol as `sse` at server level introduces any downstream issues for validation or parsers for code generation.

fmvilas:
Yeah now there's nothing in the AsyncAPI document saying that this API MUST use SSE. It only defines that "when using SSE", this message's event type is X. IMHO, there should be something else indicating that an API uses SSE, like server protocol `http+sse` or `https+sse`. I'd not allow `sse` as the server protocol because it lacks information about the HTTP counterpart, for instance, if it's secure (https) or not.

Issue #255, opened 2024-07-15
SNS section not compatible with AsyncAPI 3.0.0


mateusz-wlosek-ppro:
SNS README is not compatible with AsyncAPI 3.0.0
E.g. the section mentions `subscribe` which is not supported anymore

https://github.com/asyncapi/bindings/blob/master/sns/README.md
https://www.asyncapi.com/blog/release-notes-3.0.0

derberg:
@dpwdec @iancooper please take a look

jturek-olx:
SQS README has the same issue. 
Any news on that?

iancooper:
@jturek-olx @derberg Thanks for pinging me. I'll talk to @dpwdec

dpwdec:
@jturek-olx @derberg I'll draft a new version that uses AsyncAPI V3. However, as we know that V2 is stilled being used widely I think we should keep the V2 bindings available on master. So should we disambiguate them in same way the main project does with a `2.x.x` folder and a `3.0.0` folder?

dpwdec:
I've raised a PR for SNS: https://github.com/asyncapi/bindings/pull/260/files

iancooper:
I have merged the PR. @jturek-olx @derberg if you still have concerns please comment here and we can look at a fix

PR #261, opened 2024-09-02
feat: add AsyncAPI V3 compatible SQS binding definitions

dpwdec:
Update to SQS binding definitions for AsyncAPI V3 compatibility.

#255


derberg:
@dpwdec hey, I noticed you introduced versioning for binding. The thing is that bindings atm should be aligned with spec, which means only the latest binding version should be in `master`. We are not extending `AsyncAPI v2` any further, so should be with bindings for v2. So they should not be in `master`. In other words, current master should only store latest, which is v3

iancooper:
> @dpwdec hey, I noticed you introduced versioning for binding. The thing is that bindings atm should be aligned with spec, which means only the latest binding version should be in `master`. We are not extending `AsyncAPI v2` any further, so should be with bindings for v2. So they should not be in `master`. In other words, current master should only store latest, which is v3

Whilst we are not extending `AsyncAPI v2` any further, I don't think we necessarily want to ask folks to upgrade to V2 just to get fixes or extensions to a binding for a given provider. The upgrade to V3 might take some time for some users. I would agree that master should be V3 and that V2 should be the branch however

derberg:
I get your point. We had the same convo for Kafka binding. Afaik it was around v3 release date. Till today there were no complaints that people need updates to v2 bindings

If we keep supporting v2, people are not motivated to move to v3. As org we do all we can to motivate to migrate, converters and other tools support it already, and supported them on release date.

If in bindings we wanna go away from that model, then imho it could be a decision all binding owners agree with, so we are consistent. But my recommendation is to really consider it, be worried about it only if there is really a request coming from the community to add something to v2 binding, and that it is critical and there is no other way to do it.

Once we let in v2 support for one binding, it is opening door for precedence to do it for other bindings, and event the spec, and you know EDA and AsyncAPI maintenance is complex enough with one version in place 😅 

And in the end, if one day, we see there is a really critical reason to improve some v2 binding, then it can be done as "one-time-exception" with some dedicated branch/tag and clear story in issue/PR why we had to do it.

derberg:
@dpwdec @iancooper thoughts?

iancooper:
> @dpwdec @iancooper thoughts?

As AsyncAPI moves forward, it will have to accept that users will lag behind the latest version of the specification. Adoption has some dependency on support for earlier versions. Now, admittedly, we are towards the start of this journey, but where folks have tooling built against V2, we will break them if we don't retain compatibility with it.

So, at JET, we have tooling that uses V2. We can't shift all those folks over to V3 overnight. There are many endpoints, and even if we automated the process of converting their scripts, the review process would take weeks, and involve a lot of sign-offs. That would likely lead us to fork, so that we could maintain V2.

So I think we need a policy here. A typical one is to support the current release, and one release earlier. 

The problem then becomes, do we match bindings version to AsyncAPI version? I think for sanity yes, so we have V2 bindings for V2 and V3 for V3 etc. Now, I would caveat that. I would ensure that once V3 is out, V2 only gets additive, non-breaking changes. So you can't change the binding for V2 in a non-backwards compatible way. That is because, the only reason to have V2 is backwards compatibility, and if a new release of the middleware comes out, whilst you are still on V2, just upgrade your AsyncAPI to V3

But that also means that in V3 we might need to have breaking changes. That is because we might find that the API for middleware changes in a fashion that is not backwards compatible. We might have to use a non-SemVer approach at that point and use V3.X where X is a breaking change.

It is a slightly different painpoint than with OpenAPI, because it does not need to version.

iancooper:
BTW, I am not sure if we need to communicate binding version in the binding specification. I suspect that is mostly out-of-band, but not really thought that through @dpwdec ?

derberg:
> As AsyncAPI moves forward, it will have to accept that users will lag behind the latest version of the specification

100%, I don't think anybody really thinks people always move to latest ASAP, even though we provide proper conversion solutions.

> Now, admittedly, we are towards the start of this journey, but where folks have tooling built against V2, we will break them if we don't retain compatibility with it.

Tooling is a bit different thing. We still have tools, like parser, generator and many others that support v2 and provide bug fixes there. So nobody here is saying: stop using tools for v2 - just at the moment the message is, if you want new features that are not yet in the spec, add them to new spec or new spec binding and support this new spec feature in the tool, then just make sure you migrate to new spec. If someone can switch to using a newer binding version, why not switch to a newer spec version?

> So I think we need a policy here. A typical one is to support the current release, and one release earlier.

100% as we already had this discussion in the case of kafka for v3: https://github.com/asyncapi/bindings/issues/182
We talked about tagging there as well. Interesting is to notice that November 2023 there were worries about v2 support, slow adoption and all that stuff - 11month later, nothing bad really happened. This is why the best way for backward compatibility and any updates is to do these in a separate v2 tag - to show that any of these things are special cases and are discouraged. 

So yeah, please make sure `master` reflects `v3` and open a separate discussion where we can ping everyone, all codeowners to participate and get a consensus that we can document and follow

derberg:
@dpwdec @iancooper kind reminder

derberg:
exactly

if you really need to do some improvements in v2 lets figure out the way. We can always open a v2 branch if you really think it is needed. Although I would suggest we do it when it is really needed. As in the kafka issue I linked - we had concerns that people might want to improve v2 related binding but after so many months the reality is that nobody asked for it

dpwdec:
Hi @derberg I've updated this to be only V3 bindings.

dpwdec:
I've also incremented the version. @derberg Not sure if this should be a major version change for V3? But the other bindings don't seem to have done a major version bump.

derberg:
lgtm, yeah, for now all bindings are at version `0.x`.
lot's of work to do here in bindings - without a working group that actively sync on that topic it will be hard.

can you do the same revert in https://github.com/asyncapi/bindings/tree/master/sns ?

iancooper:
> lgtm, yeah, for now all bindings are at version `0.x`. lot's of work to do here in bindings - without a working group that actively sync on that topic it will be hard.
> 
> can you do the same revert in https://github.com/asyncapi/bindings/tree/master/sns ?

Maybe we need to establish a working group. Obviously, if folks are doing work, based on our experience of using AsyncAPI bindings, and then meet resistance to proposed changes, that is "wasted effort."

I also think that it is important to understand the AsyncAPI versioning policy, as it would make it very difficult for tool authors, both internal to organizations, and externally to support tooling, if they can't support changes for earlier versions of the specification, either to support changes to the underlying protocol used in the binding with the earlier version, or to fix issues.

We are not going to be able to debate that in this issue. But, we ought to be able to debate it somewhere.

derberg:
@dpwdec @iancooper guys please let's finish this one

Issue #254, opened 2024-06-21
ROS2 bindings


renzo-sie:
### Why do we need this improvement?

We need this improvement to seamlessly integrate ROS2, a widely-used framework in robotics, with AsyncAPI. This integration will enhance the capabilities of ROS2 by providing a standardized way to document and communicate its event-driven architecture. By including ROS2 bindings, we can ensure that ROS2 developers can leverage AsyncAPI's benefits for better interoperability, documentation, and tooling.

### How will this change help?

This change will help by providing a standardized approach to describing ROS2 communication interfaces, enabling better tooling support, improved collaboration, and more efficient development processes. It will facilitate the generation of documentation, code, and other artifacts, thus reducing the manual effort required and minimizing the risk of errors.

### How could it be implemented/designed?

We will implement the ROS2 bindings similarly to other existing bindings in AsyncAPI. This involves:

- Defining the ROS2-specific protocol and message type bindings.
- Creating documentation and examples to guide users in applying these bindings.

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

Yes I am willing to submit a PR!

renzo-sie:
Hi, is there any news on this issue? :) If I understand correctly, I still need to wait for a branch to be created to submit my code

renzo-sie:
Hi @derberg or @fmvilas , could you help with this issue?

derberg:
@renzo-sie hey, sorry for the delay. Can you provide some example AsyncAPI document with these new binding included, with comments what these binding properties are for, so it is easier to understand what you need. It will also clear out where these bindings be located, in what AsyncAPI objects, like channel or others.

Are you aware that in case of bindings, it is not enough to contribute binding, you also have to become a maintainer of it?

renzo-sie:
Hi @derberg , thanks for the response! 
Yes, I will you attached  the ROS 2 bindings README that explains what does each property do. Let me know if you have any questions. 
[README.md](https://github.com/user-attachments/files/17905656/README.md)

You can set @gramss (https://github.com/gramss) as the main mantainer for this binding,  and me as the second contact.

Would be nice to show this binding in the next ROS event in germany next week :) (https://roscon2024.de/)

[...]

PR #270, opened 2025-03-07
Initial ROS2 AsyncAPI contribution by SIEMENS AG

gramss:
**Description**

- Initial PR for ROS2 integration into AsyncAPI as introduced by #254 

Maintaining will be done through @amparo-siemens and me.
Further maintainers and improvements suggested in #254  are highly welcome.

This binding is tested already in a few internal tools and generators as seen in this presentation: [ROSCon DE](https://roscon2024.de/presentations/S4_P1__Empowering_ROS_Users_in_Industry_and_Research_with_Industrial_Automation_Equipment.pdf)

cc: @renzo-sie (thank you for the preparations!) and @fmvilas

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ This repository contains the specifications for each AsyncAPI protocol binding.
 * [NATS binding](./nats)
 * [Pulsar](./pulsar)
 * [Redis binding](./redis)
+* [ROS2](./ros2)
 * [SNS binding](./sns)
 * [Solace binding](./solace)
 * [SQS binding](./sqs)
--- a/ros2/README.md
+++ b/ros2/README.md
@@ -1,0 +1,185 @@
+# ROS 2 Bindings
+
+This document defines how to describe ROS 2-specific information in AsyncAPI.
+
+It applies to all versions of ROS 2 (foxy, galactic, humble, iron, jazzy).
+
+<a name="version"></a>
+
+## Version
+
+Current version is `0.1.0`.
+
+<a name="server"></a>
+
+## Server Binding Object
+
+This object contains information about the server representation in ROS 2. 
+ROS 2 can use either DDS or Zenoh as its middleware. 
+DDS is decentralized with no central server, so the field `host` can set to `localhost`.
+When using Zenoh, the `host` field specifies the Zenoh Router IP address.
+
+###### Fixed Fields
+
+Field Name | Type | ROS 2 Versions | Description
+---|:---:|:---:|---|
+`rmwImplementation` | string  | all | Specifies the ROS 2 middleware implementation to be used. Valid values include `rmw_fastrtps_cpp` (Fast DDS), `rmw_cyclonedds_cpp` (Cyclone DDS), `rmw_connext_cpp` (RTI Connext), and `rmw_zenoh_cpp` (Zenoh). This determines the underlying middleware implementation that handles communication.
+`domainId` | integer  | all | All ROS 2 nodes use domain ID 0 by default. To prevent interference between different groups of computers running ROS 2 on the same network, a group can be set with a unique domain ID. Must be a non-negative integer less than 232. 
+
+### Examples
+
+```yaml
+servers:
+  ros2:
+    host: localhost
+    protocol: ros2
+    protocolVersion: humble
+    bindings:
+      ros2:
+        rmwImplementation: rmw_fastrtps_cpp
+        domainId: 0
+```
+
+
+<a name="channel"></a>
+
+## Channel Binding Object
+
+A channel represents a ROS 2 topic, service or action interface.
+
+This object DOES NOT contain any ROS 2 specific properties.
+
+Example - ROS 2 topic:
+
+```yaml
+  address: /turtle1/cmd_vel
+    messages:
+      TwistMsg:
+        $ref: '#/components/messages/TwistMsg'
+```
+Example  - ROS 2 action:
+
+```yaml
+  RotateAbsoluteRequest:
+    address: /turtle1/rotate_absolute
+    messages:
+      RotateAbsoluteActionRequest:
+        $ref: '#/components/messages/RotateAbsoluteActionRequest'
+  
+  RotateAbsoluteReply:
+    address: /turtle1/rotate_absolute
+    messages:
+      RotateAbsoluteActionResult:
+        $ref: '#/components/messages/RotateAbsoluteActionResult'
+      RotateAbsoluteActionFeedback:
+        $ref: '#/components/messages/RotateAbsoluteActionFeedback'
+```
+
+<a name="operation"></a>
+
+## Operation Binding Object
+
+AsyncAPI operations with their `send` and `receive` actions map directly to ROS 2 subscribers, publishers, actions or services.
+- send -> `publisher`, `action_client`, `service_client`
+- receive -> `subscriber`, `action_server`, `service_server`
+
+###### Fixed Fields
+
+Field Name | Type | ROS 2 Versions | Description
+---|:---:|:---:|---|
+`type` | string | all | Specifies the ROS 2 type of the node for this operation. Valid values are: `publisher`, `subscriber`, `service_client`, `service_server`, `action_client`, `action_server`. This defines how the node will interact with the associated topic or action.
+`node` | string | all | The name of the ROS 2 node that implements this operation. 
+`qosPolicies` | object | all | Quality of Service (QoS) for the topic.
+
+### Quality of Service Object
+This object contains ROS 2 specific information about the Quality of Service policies.
+More information here: https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies
+
+Field Name | Type | ROS 2 Versions | Description
+---|:---:|:---:|---|
+`reliability` | string | all | One of `best_effort` or `reliable`.  More information here: [ROS 2 QoS](https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies)
+`history` | string | all | One of `keep_last`, `keep_all` or `unknown`. More information here: [ROS 2 QoS](https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies)
+`durability` | string | all | One of `transient_local` or `volatile`. More information here: [ROS 2 QoS](https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies)
+`lifespan` | integer | all | The maximum amount of time between the publishing and the reception of a message without the message being considered stale or expired. `-1` means infinite. 
+`deadline` | integer | all | The expected maximum amount of time between subsequent messages being published to a topic. `-1` means infinite.
+`liveliness` | string | all | One of `automatic`or `manual`. More information here: [ROS 2 QoS](https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies)
+`leaseDuration` | integer | all | the maximum period of time a publisher has to indicate that it is alive before the system considers it to have lost liveliness. `-1` means infinite.
+### Examples
+
+ROS 2 subscriber example:
+
+```yaml
+  receiveCmdVel:
+    action: receive
+    channel:
+      $ref: "#/channels/CmdVel"
+    bindings:
+      ros2:
+        role: subscriber
+        node: /turtlesim
+          qosPolicies:
+            history: unknown
+            reliability: reliable
+            durability: volatile
+            lifespan: -1
+            deadline: -1
+            liveliness: automatic
+            leaseDuration: -1
+```
+ROS 2 action Server example:
+
+```yaml
+  receiveRotateAbsolute:
+    action: receive
+    channel:
+      $ref: "#/channels/RotateAbsoluteRequest"
+    reply:
+      channel:
+        $ref: "#/channels/RotateAbsoluteReply"
+    bindings:
+      ros2:
+        role: action_server
+        node: /turtlesim
+```
+
+
+<a name="message"></a>
+
+## Message Binding Object
+
+ROS 2 message types (defined in .msg/.srv/.aciton files) are mapped to AsyncAPI message payloads.
+
+This object DOES NOT contain any ROS 2 specific properties.
+
+```yaml
+Vector3Msg:
+  type: object
+  properties: 
+    x:
+      type: number
+      format: double
+    y:
+      type: number
+      format: double
+    z:
+      type: number
+      format: double
+```
+
+ROS 2  Type | AsyncAPI Type | AsyncAPI Format | 
+---|:---:|---|
+bool | boolean | boolean
+byte | string | octet
+char | integer | uint8
+float32 | number | float
+float64 | number | double
+int8 | integer | int8
+uint8 | integer | uint8
+int16 | integer | int16
+uint16 | integer | uint16
+int32 | integer | int32
+uint32 | integer | uint32
+int64 | integer | int64
+uint64 | integer | uint64
+string | string | string
+array | array | --


ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+## Server Binding Object
+
+This object contains information about the server representation in ROS 2. 
+ROS 2 can use either DDS or Zenoh as its middleware. 
+DDS is decentralized with no central server, so the field `host` can set to `localhost`.
+When using Zenoh, the `host` field specifies the Zenoh Router IP address.
+
+###### Fixed Fields
+
+Field Name | Type | ROS 2 Versions | Description
```
fmvilas, line 24: I see you're including a "ROS 2 Versions" column but this column always has the "all" value. Should we get rid of it or do you expect future properties to appear only for certain versions of ROS 2?

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+Current version is `0.1.0`.
+
+<a name="server"></a>
+
+## Server Binding Object
+
+This object contains information about the server representation in ROS 2. 
+ROS 2 can use either DDS or Zenoh as its middleware. 
+DDS is decentralized with no central server, so the field `host` can set to `localhost`.
```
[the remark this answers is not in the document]
fmvilas: What about making it clear that the `host` field should be completely ignored for ROS 2?

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+        $ref: '#/components/messages/RotateAbsoluteActionRequest'
+  
+  RotateAbsoluteReply:
+    address: /turtle1/rotate_absolute
+    messages:
+      RotateAbsoluteActionResult:
+        $ref: '#/components/messages/RotateAbsoluteActionResult'
+      RotateAbsoluteActionFeedback:
+        $ref: '#/components/messages/RotateAbsoluteActionFeedback'
+```
```
fmvilas, line 76: If there's no channel binding, I recommend leaving it out of the document. E.g., the way we're doing with the AMQP 1.0 channel binding: https://github.com/asyncapi/bindings/tree/master/amqp1#channel-binding-object.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+<a name="operation"></a>
+
+## Operation Binding Object
+
+AsyncAPI operations with their `send` and `receive` actions map directly to ROS 2 subscribers, publishers, actions or services.
+- send -> `publisher`, `action_client`, `service_client`
+- receive -> `subscriber`, `action_server`, `service_server`
```
fmvilas, line 84: I'm not sure I understand this. Could we improve this description?

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+###### Fixed Fields
+
+Field Name | Type | ROS 2 Versions | Description
+---|:---:|:---:|---|
+`type` | string | all | Specifies the ROS 2 type of the node for this operation. Valid values are: `publisher`, `subscriber`, `service_client`, `service_server`, `action_client`, `action_server`. This defines how the node will interact with the associated topic or action.
```
fmvilas, line 90: So, if I understand correctly, when the `action` field is `send`, this value can be `publisher`, `action_client`, or `service_client`. And when `action` is `receive`, this value can be `subscriber`, `action_server`,  or `service_server`. Is that right? If so, this relationship should be expressed here instead of above.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Operation Binding Object
+
+AsyncAPI operations with their `send` and `receive` actions map directly to ROS 2 subscribers, publishers, actions or services.
+- send -> `publisher`, `action_client`, `service_client`
+- receive -> `subscriber`, `action_server`, `service_server`
+
+###### Fixed Fields
+
+Field Name | Type | ROS 2 Versions | Description
```
fmvilas, line 88: Same as above with the ROS 2 Versions column.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+---|:---:|:---:|---|
+`type` | string | all | Specifies the ROS 2 type of the node for this operation. Valid values are: `publisher`, `subscriber`, `service_client`, `service_server`, `action_client`, `action_server`. This defines how the node will interact with the associated topic or action.
+`node` | string | all | The name of the ROS 2 node that implements this operation. 
+`qosPolicies` | object | all | Quality of Service (QoS) for the topic.
```
fmvilas, line 92:
```suggestion
`qosPolicies` | [Quality of Service Policy Object](#QoSPolicyObject) | all | Quality of Service (QoS) for the topic.
```

Please note I'm adding the `#QoSPolicyObject` anchor below in another comment.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Quality of Service Object
```
fmvilas, line 94:
```suggestion
<a name="QoSPolicyObject" />

### Quality of Service Object
```

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+This object contains ROS 2 specific information about the Quality of Service policies.
+More information here: https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies
+
+Field Name | Type | ROS 2 Versions | Description
```
fmvilas, line 98: Same as above with the ROS 2 Versions column.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+---|:---:|:---:|---|
+`reliability` | string | all | One of `best_effort` or `reliable`.  More information here: [ROS 2 QoS](https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies)
+`history` | string | all | One of `keep_last`, `keep_all` or `unknown`. More information here: [ROS 2 QoS](https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies)
+`durability` | string | all | One of `transient_local` or `volatile`. More information here: [ROS 2 QoS](https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies)
+`lifespan` | integer | all | The maximum amount of time between the publishing and the reception of a message without the message being considered stale or expired. `-1` means infinite. 
+`deadline` | integer | all | The expected maximum amount of time between subsequent messages being published to a topic. `-1` means infinite.
+`liveliness` | string | all | One of `automatic`or `manual`. More information here: [ROS 2 QoS](https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies)
```
fmvilas, line 105:
```suggestion
`liveliness` | string | all | One of `automatic` or `manual`. More information here: [ROS 2 QoS](https://docs.ros.org/en/jazzy/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-policies)
```

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+`leaseDuration` | integer | all | the maximum period of time a publisher has to indicate that it is alive before the system considers it to have lost liveliness. `-1` means infinite.
```
fmvilas, line 106:
```suggestion
`leaseDuration` | integer | all | The maximum period of time a publisher has to indicate that it is alive before the system considers it to have lost liveliness. `-1` means infinite.
```

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+ROS 2 subscriber example:
+
+```yaml
+  receiveCmdVel:
+    action: receive
+    channel:
+      $ref: "#/channels/CmdVel"
+    bindings:
+      ros2:
+        role: subscriber
```
fmvilas, line 118: It would be great to make this value the default so people don't have to specify it.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+      $ref: "#/channels/RotateAbsoluteRequest"
+    reply:
+      channel:
+        $ref: "#/channels/RotateAbsoluteReply"
+    bindings:
+      ros2:
+        role: action_server
+        node: /turtlesim
+```
+
```
fmvilas, line 144: Can we get examples for the other roles too? Especially, when the `action` is `send`.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+<a name="message"></a>
+
+## Message Binding Object
+
+ROS 2 message types (defined in .msg/.srv/.aciton files) are mapped to AsyncAPI message payloads.
```
fmvilas, line 150:
```suggestion
ROS 2 message types (defined in .msg/.srv/.action files) are mapped to AsyncAPI message payloads.
```

fmvilas: What do you mean by "are mapped"?

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+    x:
+      type: number
+      format: double
+    y:
+      type: number
+      format: double
+    z:
+      type: number
+      format: double
+```
```
fmvilas, line 167: If this is not getting any specific configuration for ROS 2, we should leave it out to avoid confusions.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+ROS 2  Type | AsyncAPI Type | AsyncAPI Format |
```
fmvilas, line 169:
```suggestion
ROS 2 Type | AsyncAPI Type | AsyncAPI Format | 
```

fmvilas: What's this table about? When should it be used and what for?

fmvilas (changes requested):
Left a round of review. Thanks for putting it together, Florian 🙏

ros2/README.md
```diff
@@ -0,0 +1,185 @@
+# ROS 2 Bindings
+
+This document defines how to describe ROS 2-specific information in AsyncAPI.
+
+It applies to all versions of ROS 2 (foxy, galactic, humble, iron, jazzy).
```
Achllle, line 5:
```suggestion
It applies to all [distributions of ROS 2](https://docs.ros.org/en/rolling/Releases.html).
```
this list will immediately become out of date, assuming that compliance with newer ROS 2 distros is automatic

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+This object contains information about the server representation in ROS 2. 
+ROS 2 can use either DDS or Zenoh as its middleware. 
+DDS is decentralized with no central server, so the field `host` can set to `localhost`.
+When using Zenoh, the `host` field specifies the Zenoh Router IP address.
+
+###### Fixed Fields
+
+Field Name | Type | ROS 2 Versions | Description
+---|:---:|:---:|---|
+`rmwImplementation` | string  | all | Specifies the ROS 2 middleware implementation to be used. Valid values include `rmw_fastrtps_cpp` (Fast DDS), `rmw_cyclonedds_cpp` (Cyclone DDS), `rmw_connext_cpp` (RTI Connext), and `rmw_zenoh_cpp` (Zenoh). This determines the underlying middleware implementation that handles communication.
```
Achllle, line 26: I would again link to the [RMW page](https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Different-Middleware-Vendors.html) rather than list the options.

Achllle (commented):
My 2c as a ROS dev: ROS 2 parameters are a must-have for this to represent a ROS 2 interface.

ros2/README.md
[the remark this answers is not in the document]
amparo-siemens: It is not the case if the middleware is Zenoh since the host will be used to specify the Router IP address

ros2/README.md
amparo-siemens: Agree. We'll do that

ros2/README.md
amparo-siemens: Agree. We'll do that

ros2/README.md
amparo-siemens: It will be put in the description field of the table

ros2/README.md
amparo-siemens: Yes, we will put the explanation here

ros2/README.md
amparo-siemens: Yes

ros2/README.md
amparo-siemens: In my opinion does not make sense since in ROS2 there is no default value for roles. It depends if you are using topics, actions, services... and it depends if you are publishing, subscribing... But we are completely open to suggestions. @Achllle what is your opinion here?

ros2/README.md
amparo-siemens: Sure, we wiil do that

ros2/README.md
amparo-siemens: Explained in the last comment

ros2/README.md
amparo-siemens: Agree

ros2/README.md
amparo-siemens: Agree

ros2/README.md
amparo-siemens: Agree

Achllle:
Considering the ROS 2 version of this effort, [node_idl / NoDL includes them in their spec](https://github.com/ubuntu-robotics/design/blob/node_idl/articles/ros2_nodl.md#parameter), I would argue this is a must-have for a complete spec. It's fine by me if it happens in two phases, but the point of a spec is that it's complete.

ros2/README.md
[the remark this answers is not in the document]
gramss: _TL;DR_: it depends.. There are a lot of valid and in the wild use cases of a `host` address that would make sense to have documented in an AsyncAPI representation.
---
_Long version_:
If you don't know it already: you need to see ROS as an abstraction layer of an actual communication layer.
The RMW (ROS MiddleWare) Layer makes sure to map the ROS communication semantic like topic (pub/sub), services or actions to the respecting transport protocol. The ROS user itself should not notice such things as long as the RMW is consistent.

The standard for ROS2 was long DDS. And those DDS systems only need a `host` aka discovery service in very specific configurations (aka not default). Zenoh is currently a new promising middleware that will come out of the box with a `host` IP that will do the discovery of the nodes and topics. Zenoh is currently not a default for ROS2, but very anticipated by the community.

If you want to get really creative, you could even use email as the transport layer.. But this is probably a 0,00001% relevance but makes the flexibility of the RMW pretty clear.

Hope that helped to understand some ROS communication internals.

ros2/README.md
gramss: hm.. Just had time to dig a little bit deeper into channels.

Would you describe this picture here as a channel?
image
[Source from OSRF](https://osrf.github.io/ros2multirobotbook/ros2_design_patterns.html)

Or another example here (circle = node == application and box = topic == message)


The message `/joint_states` is consumed by 2 nodes..

For me, `topic` translates to an asyncAPI `message` and a ros `node` translates to an  asyncAPI `application`.
There can be one-to-many communications in ROS between one node (pub) and many other nodes (subs). Same applies for the other ros2 `roles`.

But I am unsure how this can be represented as a channel here. The specifics how the message is transported to many clients is defined inside of the concrete rmw implementation(?). A node can only tune how to react to a new event but not influence the transport mode other than specifying the QoS level.

Does @Achllle maybe has an idea about that? Here is the asyncAPI definition: [link](https://www.asyncapi.com/docs/concepts/channel)

I am having a hard time especially with translating this sentence from the doc to ROS:

> The producer sends a message through the channel, which then queues the message for delivery to the appropriate consumers.

gramss:
> Considering the ROS 2 version of this effort, [node_idl / NoDL includes them in their spec](https://github.com/ubuntu-robotics/design/blob/node_idl/articles/ros2_nodl.md#parameter), I would argue this is a must-have for a complete spec. It's fine by me if it happens in two phases, but the point of a spec is that it's complete.

yes. I agree that parameters need to be added to the ros2 bindings. In our use cases,  we did not use them so far. That is why they slipped our implementation. I would like to add them in a second step.

From what I've [gathered](https://docs.ros.org/en/rolling/Concepts/Basic/About-Parameters.html#parameter-types), parameters are interfaced through fixed services that all nodes have, unless pro-actively deactivated. What can be transported is also fixed in terms of data / message types. So like you put it, it might only be necessary to add to a node==application what parameters it has and probably not list all the additional `service` interfaces that a node has under the hood.

We need to investigate this more. I still am having trouble how it embeds into an asyncAPI binding.. Is somebody in the ros-infrastructure team already familiar with asyncAPI @Achllle ?

Achllle:
The best way to get involvement from the ROS community would be via a post in ROS discourse!

ros2/README.md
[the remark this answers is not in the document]
fmvilas: Thanks for the explanation folks. Then I'd try to make it clear that, in certain cases, the host should simply be ignored. For instance:

```suggestion
When using DDS, the field `host` MUST be ignored since DDS is decentralized and, therefore, has no central server.
```

Would that make sense? Please note that in the below line we're specifying that Zenoh makes use of the host field so I'm clarifying what should be the case for DDS. Only DDS. I'm trying to get as specific as possible but you folks are the ROS2 experts here so please let me know if I'm dropping nonsense 😅

ros2/README.md
fmvilas: mmm 🤔  from what I'm reading [here](https://docs.ros.org/en/foxy/Concepts.html#graph-concepts), it seems that AsyncAPI channels map to ROS2 topics and AsyncAPI messages map to ROS2 messages (or data types, data units). So, from what I'm learning so far:

* ROS2 node -> AsyncAPI application (i.e., the whole AsyncAPI file)
* ROS2 middleware (DDS, Zenoh) -> AsyncAPI server.
* ROS2 topic -> AsyncAPI channel
* ROS2 message/data type -> AsyncAPI message

Feel free to correct me as I'm looking at this for the first time.

fmvilas: From the docs [here](https://docs.ros.org/en/foxy/Tutorials/Beginner-Client-Libraries/Writing-A-Simple-Py-Publisher-And-Subscriber.html#write-the-subscriber-node):

```python
        self.subscription = self.create_subscription(
            String,
            'topic', # <-- This is the AsyncAPI's channel address
            self.listener_callback,
            10)
        self.subscription  # prevent unused variable warning

    def listener_callback(self, msg): # <-- `msg` is the AsyncAPI's message
        self.get_logger().info('I heard: "%s"' % msg.data) # <-- msg.data is the AsyncAPI's message payload
```

And I'm assuming, in `listener_callback`, the `msg` variable will contain a `msg.header` which maps directly to AsyncAPI's message headers.

ros2/README.md
fmvilas: Sure, I'm all in with having it, I just think we have to explain what is for. Right now it's just there without any further explanation.

fmvilas: Sure, I'm all in with having it, I just think we have to explain what is for. Right now it's just there without any further explanation.

ros2/README.md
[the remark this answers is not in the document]
Achllle: Technically `none` would be more accurate. ROS 2 is able to communicate across networks and you can even specify which nodes (producers, consumers, applications) can run on which machine _at runtime_.

ros2/README.md
Achllle: @fmvilas based on my exp with ROS 2 and understanding of the AsyncAPI glossary, your assessment seems correct.

> msg variable will contain a msg.header

yes, though keep in mind:

1. there is an 'official' [`std_msgs/Header.msg`](https://docs.ros2.org/foxy/api/std_msgs/msg/Header.html) but it's not in all message definitions per se ([example](https://docs.ros.org/en/noetic/api/std_msgs/html/msg/Float32.html))
2. furthermore, someone can very well create a field `header` in their message and proceed to put the type as `std_msgs/Float32` or whatever they want. 

All that to say that that mapping probably shouldn't be enforced or separately implemented. I don't think we want AsyncAPI's header to vary based on the message.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+`domainId` | integer  | all | All ROS 2 nodes use domain ID 0 by default. To prevent interference between different groups of computers running ROS 2 on the same network, a group can be set with a unique domain ID. Must be a non-negative integer less than 232.
```
Achllle, line 27:
```suggestion
`domainId` | integer  | all | All ROS 2 nodes use domain ID 0 by default. To prevent interference between different groups of computers running ROS 2 on the same network, a group can be set with a unique domain ID. [Must be a non-negative integer less than 232](https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Domain-ID.html).
```

> domain IDs 0-101 and 215-232 can be safely used without colliding with ephemeral ports. 

This is why I would add the link

ros2/README.md
Achllle: Not sure if I'm nitpicking, but technically speaking:

```suggestion
AsyncAPI operations with their `send` and `receive` actions map directly to ROS 2 subscribers, publishers, actions or services.
- producer -> `publisher`, `action_client`, `service_client`
- consumer -> `subscriber`, `action_server`, `service_server`
```

send and receive refers to publisher.publish(), receive would point to a callback()/register_callback()

ros2/README.md
Achllle: I do believe this belongs here, agreed with adding explanation. Suggest including a link to the [design document](https://design.ros2.org/articles/legacy_interface_definition.html)

ros2/README.md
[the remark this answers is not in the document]
fmvilas: `none` is fine too. As long as we're talking about the `none` string because the `host` field is required so it can't be empty.

ros2/README.md
Achllle: > Is header a field of the data/payload?

Correct! [See an example](https://docs.ros.org/en/ros2_packages/rolling/api/point_cloud_interfaces/msg/CompressedPointCloud2.html). We should not specify this as part of the AsyncAPI  binding. ROS often suggests a best practice, but it's not enforced since robotics is so diverse in its applications. The underlying middleware implementation might have a header, but since that depends on the implementation, I don't think it's worth trying to map this.

fmvilas: I see. Yeah, that makes sense then.

Now this opens another question to me: in the binding, should we say that the message payload must not contain the header field and, instead, it should be defined in the message headers? Because, what happens if someone defines message headers and a different `header` field in the message payload in AsyncAPI? Who's right in that case? 🤔

Achllle: No, putting a header message is essential for some applications in ROS. A ROS [std_msgs/Header](https://docs.ros2.org/foxy/api/std_msgs/msg/Header.html) contains a `frame_id`, used to represent a coordinate frame the data in the message applies to. 
I would really treat the header in the ROS message payload as an unfortunate naming collision. With that in mind, I think that means we'll just have to have the two types of headers. My understanding is that for this binding, the AsyncAPI headers just wouldn't apply. That or some underlying DDS/Zenoh/RMW header should be used. I'm not familiar enough with that lower level to comment with authority.

fmvilas: I would put some text in the binding indicating that the `headers` field on the Message Object will be ignored. Instead, implementations should use the `header` field of the payload as "the headers".

amparo-siemens: > mmm 🤔 from what I'm reading [here](https://docs.ros.org/en/foxy/Concepts.html#graph-concepts), it seems that AsyncAPI channels map to ROS2 topics and AsyncAPI messages map to ROS2 messages (or data types, data units). So, from what I'm learning so far:
> 
> * ROS2 node -> AsyncAPI application (i.e., the whole AsyncAPI file)
> * ROS2 middleware (DDS, Zenoh) -> AsyncAPI server.
> * ROS2 topic -> AsyncAPI channel
> * ROS2 message/data type -> AsyncAPI message
> 
> Feel free to correct me as I'm looking at this for the first time.

This is how I undurstand the mapping between ROS and AsyncAPI

RosMapToAsyncAPI

amparo-siemens: > I would put some text in the binding indicating that the `headers` field on the Message Object will be ignored. Instead, implementations should use the `header` field of the payload as "the headers".

What do you think about this explanation in the message binding?

## Message Binding Object

While this object DOES NOT contain any ROS 2 specific properties, it is important to understand how to express the different [ROS2 data types](https://design.ros2.org/articles/legacy_interface_definition.html#:~:text=to%20IDL%20types-,ROS%20type,string,-The%20mapping%20of) in [AsyncAPI message types and formats](https://www.asyncapi.com/docs/reference/specification/v3.0.0#dataTypeFormat:~:text=The%20formats%20defined%20by%20the%20AsyncAPI%20Specification%20are%3A).

ROS 2  Type | AsyncAPI Type | AsyncAPI Format | 
---|:---:|---|
bool | boolean | boolean
byte | string | octet
char | integer | uint8
float32 | number | float
float64 | number | double
int8 | integer | int8
uint8 | integer | uint8
int16 | integer | int16
uint16 | integer | uint16
int32 | integer | int32
uint32 | integer | uint32
int64 | integer | int64
uint64 | integer | uint64
string | string | string
array | array | --

It is important to understand that the message header of the AsyncAPI specification does not map with the message header in ROS2.
For this reason, the AsyncAPI message header will be ignored. If you want to define the header of a message in ROS2, you should put in the payload of the message. [Example](https://docs.ros.org/en/ros2_packages/rolling/api/point_cloud_interfaces/msg/CompressedPointCloud2.html):

```yaml
channels:
  Twist:
    address: /pointCloud
    messages:
      CompressedPointCloud2:
        $ref: "#/components/messages/CompressedPointCloud2"
components:
  CompressedPointCloud2:
    tags:
      - name: msg
    payload:
      type: object
      properties:
        header:
          $ref: "#/components/messages/std_msgs/header/payload"
        height:
          $ref: "#/components/messages/uint32/payload"
        width:
          $ref: "#/components/messages/uint32/payload"
        fields:
          $ref: "#/components/messages/sensor_msgs/PointField/payload"
        is_bigendian:
          $ref: "#/components/messages/bool/payload"
        point_step:
          $ref: "#/components/messages/uint32/payload"
        row_step:
          $ref: "#/components/messages/uint32/payload"
        compressed_data:
          $ref: "#/components/messages/uint8/payload"
        is_dense:
          $ref: "#/components/messages/bool/payload"
        format:
          $ref: "#/components/messages/string/payload"     
```

ros2/README.md
amparo-siemens: The reason that we used send/receive is because those are the [only options](https://www.asyncapi.com/docs/reference/specification/v3.0.0#messagesObject:~:text=DESCRIPTION-,action,-%22send%22%20%7C) for the field `action` in AsyncAPI and we were tying to map it to the different ROS2 types, since we cannot change the valid options for the field `action`:

ros2/README.md
amparo-siemens: The idea is to have a ros2 binding named as "role" inside the operation bindings to define if a node is publisher/subscriber or service/action client/server. Right now you don't see it in the binding documentation because it does not exist yet. What we are trying here is to define the necessary bindings to put them as official bindings (in the same way that you see right know the mqtt ones) and then you will be able to see them. Right know all ros2 bindings that don't use the existing ones, are experimental.
Summarizing, in bindings we can get "anything" we want.

ros2/README.md
fmvilas: @amparo-siemens so you think the AsyncAPI operation maps the ROS 2 node? I'm not quite sure I follow.

amparo-siemens: Theoretically it could be as you said: ROS2 node -> AsyncAPI application (i.e., the whole AsyncAPI file)
It depends if we want multiple nodes in multiple asyncAPI files or one system with multiple nodes in one asyncAPI.

We as Siemens prefer the second. So you have the whole system with multiple nodes in one file. In that file you have the different topics running in that system (channels) with the definition of those interfaces (messages) and the nodes that are using those topics are specified in the operations:

Operation definition: [_In an AsyncAPI document, operations describe your application's behaviors and capabilities by exchanging messages through channels configured in the AsyncAPI document._](https://www.asyncapi.com/docs/concepts/asyncapi-document/adding-operations#:~:text=In%20an%20AsyncAPI%20document%2C%20operations%20describe%20your%20application%27s%20behaviors%20and%20capabilities%20by%20exchanging%20messages%20through%20channels%20configured%20in%20the%20AsyncAPI%20document.)

Node definition: [_Each node can send and receive data from other nodes via topics, services, actions, or parameters._](https://docs.ros.org/en/rolling/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Nodes/Understanding-ROS2-Nodes.html#:~:text=Each%20node%20can%20send%20and%20receive%20data%20from%20other%20nodes%20via%20topics%2C%20services%2C%20actions%2C%20or%20parameters.)

As you see in the next image, this does not mean that one node is one opeation. Since the same node could be a publisher in one topic (send in one channel) and a Service Server in other topic (receive in this other channel).

RosMapToAsyncAPI_2

So one node could have several operations.

fmvilas: > Theoretically it could be as you said: ROS2 node -> AsyncAPI application (i.e., the whole AsyncAPI file)
It depends if we want multiple nodes in multiple asyncAPI files or one system with multiple nodes in one asyncAPI.
>
> We as Siemens prefer the second. So you have the whole system with multiple nodes in one file.

I see. Just for reference, we often recommend that an AsyncAPI file represents a single application/program/node. For systems, you should combine multiple AsyncAPI files. I understand the convenience of having everything in a single file but just wanted to give a heads up that this would go against the philosophy of AsyncAPI. You may encounter issues down the track because of the misalignment of both approaches.

> As you see in the next image, this does not mean that one node is one opeation. Since the same node could be a publisher in one topic (send in one channel) and a Service Server in other topic (receive in this other channel).

Precisely. That's why I think the AsyncAPI file should represent a node. Because it can have multiple operations (`operations` in AsyncAPI). If you make the AsyncAPI file represent a group of nodes (a system) and, for instance, want to generate code, how do you map each operation to each node? How do you know which operation belongs in each node?

Achllle: A ROS 2 API typically is the API of the ROS 2 _package_. It's very possible for a package to have dozens of nodes. Dozens of of AsyncAPI files that would need to be combined feels cumbersome. Often an end-user of the package doesn't 'care' which node does what, but instead which topics/services/actions are available.
This is further complicated by the existence of [composable nodes](https://docs.ros.org/en/jazzy/Tutorials/Intermediate/Composition.html) where nodes can be composed at runtime.

fmvilas: Alright, I see a misalignment in the philosophy of both technologies. Whatever we come up with here, let's try to make sure it makes sense for both sides.

> it could be possible to have like a top file for the system contaning all these AsyncAPI files of each node?

Totally. Even what you suggested initially is possible. I was just discouraging it because I saw some inconsistencies. However... 👇 

> Often an end-user of the package doesn't 'care' which node does what, but instead which topics/services/actions are available.

If that's usually the case, I can see it may be unnecessary to have multiple files.

We've seen this use case before. For instance, people who want to use AsyncAPI to offer a sort of a "menu" of the available channels and their messages. Often to configure a broker. That's been the case for IBM folks, for instance.

I started a discussion about this a few years ago: https://github.com/asyncapi/spec/issues/628. This derived in the following discussion: https://github.com/asyncapi/spec/issues/658. The result of all these discussions was that we're not planning to implement support for defining multiple applications in a single AsyncAPI file. That said, "not planning" doesn't mean we're against it. In fact, most of us agreed that there should be some kind of mechanism for that (maybe a parent spec).

In the meantime, if you folks truly believe that the best way forward is to use AsyncAPI to define a collection of nodes, go ahead and make it clear in the ROS 2 binding. That would somehow clash with philosophy of other bindings but would set a precedent to reopen the discussions I shared above.

An idea that comes to my mind is that you could leverage the Server Binding to specify which behavior you want to stick to: defining a node or defining a collection of nodes. This way, you give developers the possibility to do both things while making it explicit. That would not go against the spec because, as stated in the [Application definition](https://www.asyncapi.com/docs/reference/specification/v3.0.0#definitionsApplication) of the spec: "An application is any kind of computer program **or a group of them**."

So long we make it clear, I think it will be fine.

gramss: off topic:
okay, thank you all for pushing the topic forward! 👍 

I agree on all the above points, thank you @amparo-siemens for representing our groups view.

----
Regarding the open question here for multiple asyncAPI files for a system.

> "An application is any kind of computer program or a group of them."

The nucleus is that you have one packages with one node and one operation (topic/service/action).
In the case of [Navigation2](https://github.com/ros-navigation/navigation2) you have ~36 package and ~70-100 nodes (depending on your config of that project for your specific use case).

For me my view on this boils down to: **Why** and **when** do I need an asyncAPI documentation and for **what**.

**Why**: because I want to look up how things work and/or what a specific thing (= GitHub project or asyncAPI application) offers

**When**: before I download the application and add it to my robot OR when I have a running composition of multiple applications

**What**: As a user I do not care which package offers which functionality. There are even packages that do offer no operations (e.g. [nav2_msgs](https://github.com/ros-navigation/navigation2/tree/main/nav2_msgs)) As a user I care about what operations are offered to me and what their configuration currently is (e.g. network and parameters (currently missing))

### Summary

I believe that we have two overarching use cases for asyncAPI with ROS2.

1. before I download applications (they can contain out of multiple packages, but I only care about the operations and channels offered)
2. during runtime of my application (=robot with n-packages & n-nodes & n-operations & n-channels)

**That boils down for me to have always a minimum of asyncAPI files.**

A feature to group asyncAPI files in a viewer for use case 1 in the scouting phase is of course nice.

For use case 2 we at Siemens already have an introspection tool in the pipeline where we want to gather everything running into one dedicated asyncAPI file.

In both use cases I do not see a benefit to have more individual asyncAPI files

> In the meantime, if you folks truly believe that the best way forward is to use AsyncAPI to define a collection of nodes, go ahead and make it clear in the ROS 2 binding

I truly believe so, yes. A robot/application with ROS2 is in 99% of the time a composition of nodes and I want to look at the whole thing from different angles.
Take a look also at the ros2 basic example, the turtlesim (also used in Nav2 for navigation example and mascot of ROS btw.).



This QT based RosGraph tool is the default way to look at complex ROS2 systems. But what is truly missing is a way to dig into what the channels are really transporting. I think asyncAPI together with the ROS2 introspection capabilities would be a really good addition to such views to hold the important information in one place. And for that I think we need a combined asyncAPI that hosts the complete application and not jump from one package to the other to find "the one service to fix all of my problems"™ 🙂

Achllle: > As a user I do not care which package offers which functionality.

I disagree with this. I would say nav2 is one of the largest projects in ROS out there. It should be possible for someone to use just a few packages from the stack but not all. So now a user would have to go figure out which topic/service/action is available in which package. When you move to commercial robot applications, nodes are logically grouped within packages which should be able to stand alone (but including interface packages). The packages often offer very different functionalities that have little to do with one another. This modularity is a core tenet in ROS, and interface description should reflect that. Additionally, such robots often use the entire nav2 stack, so now you're already dealing with at least two AsyncAPI descriptions for one robot. These robots also often have thousands of topics. If the API description is not logically grouped per package, it will in fact become harder to find what you're looking for. Finally, keeping the API descriptions inside the respective packages themselves makes maintenance less prone to 'adding a topic and forgetting to add it to the API description'.

Given you can easily combine AsyncAPI files, you can readily achieve a single interface which addresses all of your reasons for wanting a single interface description.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+This object DOES NOT contain any ROS 2 specific properties.
+
+Example - ROS 2 topic:
+
+```yaml
+  address: /turtle1/cmd_vel
+    messages:
+      TwistMsg:
+        $ref: '#/components/messages/TwistMsg'
```
Achllle, line 58: This seems like a custom message type which is confusing since the [official page](https://docs.ros.org/en/crystal/Tutorials/Topics/Understanding-ROS2-Topics.html#ros2-topic-list) uses [geometry_msgs/Twist](https://docs.ros.org/en/rolling/p/geometry_msgs/msg/Twist.html)

This also begs the question: when an API uses a message definition that is part of the official ROS 2 distribution or from a package on rosdistro (e.g. geometry_msgs, nav_msgs), is there a way we can avoid needing to describe that message in the API declaration? That seems error prone and overly verbose/cluttering.

ros2/README.md
gramss: @Achllle these are some really good points and gets me out of our bubble.. And yes, I choose nav2 as an example to show the "upper" limits what this binding should be able to deal with.

After reading your comment more throughly, I can only say that I fully agree with you.
My idea was solely to maybe reduce AsyncAPI files per GitHub project, but you got a point that it is easier to determine the content of a package if the AsyncAPI about that specific package just also sits in that package folder (regardless if it is an independent GitHub project or a sub-package of a larger project).

And yes, we 100% need to group multiple AsyncAPI files (meaning 2 till n) into a robotic application.
In the past I did discuss this a lot with @amparo-siemens why we would need that if we can generate everything out of runtime, but we totally missed the point from @Achllle in our discussions, so thank you for that!

Still, I see 2 use cases. "Green field" where you have all the source code and "brown field" where you want to get as much information form a closed-source or legacy system. They are the same as above and I gave more examples down below as-well. (and have them part of my ROSConDE talk)

---

Short explanation of "our bubble": We already have a PoC of an introspection tool that creates an AsyncAPI file from runtime ros2 systems. But this only works on Jazzy+ with limitations on services and actions (we need the source files here). https://github.com/ros2/rmw/issues/356 & https://github.com/ros2/ros2cli/pull/846

**So probably a good way in the middle is to attach an AsyncAPI file per package, that holdes at least one node.**
As you said, combining them goes always, but having them at their packages helps finding the implementation of the documented interface faster.

Where I don't see it fit are interface packages = e.g. nav2_msgs packages that do not hold any nodes = no AsyncAPI application = no AsyncAPI file? @amparo-siemens could you give here a short insight how we use the .msg definitions to build up an AsyncAPI representing an application? That could shine a light on why I think we do not need an AsyncAPI representation in those sole msgs-packages that are quite common in the ROS universe (are they are REP?).

Also, we cannot go deeper, as a per node representation is hard to do code wise with the different rcl implementations? Somehow I would love to jump to the exact definition of a publisher from the AsyncAPI doc into the code.. Maybe something to digest later.

---
And now I think I can navigate our discrepancy:

Your view is for how to get proper documentation for per-package documentation. Your points hold true (and I totally agree to them) for:
- share code online
- reuse modular code in-house
- general good code etiquette 

I am coming more from a use case where I want to have a tailored documentation for my robot and come from scenarios like:
- dealing with legacy/3rd-party code and do not want to or cannot read the code (unless necessary) 
- missing / non-standard /scattered documentation
- only binary of code available, no source code
- no direct access to robot machine (only ros2 reachable via network)
- I want to generate a tailor made documentation on the fly (robot node orchestration/composition changes on the fly, so should documentation)

### What does this mean for this discussion?

When the source code is available = have an AsyncAPI file per package.
If you just have a blackbox (brownfield) then on AsyncAPI file is fine. Or at least where you can roughly group them into more separate applications (sensory, drive, robot arm, ...)? 

Which means we can stay with the original philosophy from @fmvilas of AsyncAPI ? The only downside is that a ros2-package = application can still have n-nodes with each n-operators.

Achllle: Appreciate the context! I agree we should let the user decide based on their goals and not impose any particular format. 

Regarding interface-definition only packages, I'd like to add this [recent comment](https://github.com/asyncapi/bindings/pull/270#discussion_r2021867350) I made into the mix. Ideally a user is able to specify the use of a common or even custom message definition without having to specify each time what's in that message. So what I'd expect is that indeed a `my_custom_interfaces` package has a AsyncAPI yaml file that just contains the `messages` definitions and another yaml file in another package is able to reference those. @fmvilas is this something that's achievable with the spec?

amparo-siemens: If I understood you correctly, this is how we are working right now.

image

We have a header file that references the different interfaces files:
image
image

gramss: @Achllle appreciate your time understanding our context! 🤗 

I am unsure, but I have a hard time to see more than those two use cases (green field & brown field). Even if they are mixed (legacy / black box code that you want to extent with source code), you can clearly separate the mix.

So I would suggest to at least have a best effort guidelines for those use cases:

> When the source code is available = have an AsyncAPI file per package.
If you just have a blackbox (brownfield) then one AsyncAPI file is fine. Or at least where you can roughly group them into more separate applications (sensory, drive, robot arm, ...)?

---

Now to the interface-only packages.
Thank you @amparo-siemens for providing a bit more insights!

Again coming from the introspection use case, I am hugely in favour of generating the documentation on the fly.
During the last decade in ROS (I seem to be getting old 🥲), I saw quite some changes in the 

> use of a common [interfaces]

E.g. the joystick message type which resorted some array elements which resulted in a broken robot in my lab while upgrading from kinetic to noetic. Or the dropping of std_msgs/Bool etc.

There so often happens something and right now those messages are human identified by their name but really need to be identified by their hashes.
Having ROS2 also be flexible on non-hash compatible messages does not really help here.
(Unsure if this is by design or an issue that should be reported as I am writing this..)

```mermaid
sequenceDiagram
    participant NodeA
    participant NodeB

    NodeA->>NodeB: Message my_custom_Bool (Bool, Float)
    Note right of NodeB: Only receives Bool
```
explaination of the diagram:

> Node A has a message type my_custom_Bool and sends a message.
> Node B has a message type my_custom_Bool and receives this message.
> 
> But Node A has the message type defined to hold a Bool and a Float.
> While Node B only has the message type defined to hold a Bool
> Both are build in different workspaces but communicate through ROS2 non-the-less

@renzo-sie did here some tests 1 or 2 years ago. Could you shortly clarify that my memory of our tests checks out here? thx

---
### Comming back to how I would approach interface-only packages

As one has access to the interface information in green field and brown field (Jazzy+) I see no point in having a static "commodity" database to get this information from. Maybe a user feels like it to twist/re-introduce an old std_msgs/Bool and add a `float` while at it.. The name of the interface stays the same, you really need to look into the hashes and then the source files to check what really is inside right now.

And while you are at it looking at all the files anyways to check them against some database just to be sure that everything checks out.. It is faster to just take the "live" data from the static (greenfield) or runtime (brownfield) environment in my honest opinion.

ros2/README.md
gramss: Would you mind rephrasing your question here, please?
> This seems like a custom message type

Do you mean the representation in the AsyncAPI file and the look and feel of it?
Or do you mean the concrete example?

If you mean the latter (`geometry_msgs/Twist` vs `TwistMsg`) then we need to ask @amparo-siemens .
Could you please check if this is just a typo or has something to do how we create the references to messages.
Personally, I believe this is just a typo and agree to @Achllle that this needs to be fixed. Thanks for pointing it out!

**We definitely want to stay with the official turtlesim example!**

Regarding your question, please check out my [answer here](https://github.com/asyncapi/bindings/pull/270#discussion_r2022314449) that vouches against the use of e.g. a "database" of common messages. Always bearing in mind not to have to write the AsyncAPI files / message definitions yourself/manually..

gramss:
> The best way to get involvement from the ROS community would be via a post in ROS discourse! @Achllle 

https://discourse.ros.org/t/invitation-to-collaborate-on-asyncapi-specification-for-ros2/42915
Finally came around to get the discourse post together.

@fmvilas if you have the time, it would be a pleasure to have you join such a meeting.
@amparo-siemens and myself try to bring a short summary up until the meeting about the status of the ongoing discussion here.

ros2/README.md
gavanderhoorn: Came here via the Discourse post, and this comment caught my eye:

> But Node A has the message type defined to hold a Bool and a Float.
> While Node B only has the message type defined to hold a Bool
> Both are build in different workspaces but communicate through ROS2 non-the-less

this has been a big issue in ROS 2 up until Iron: that's when msg content hashing was implemented generally -- similar to how ROS 1 had this since almost the beginning.

I'm not dismissing your experience, just thought I'd point out this should not be possible anymore in Jazzy and newer.

> There so often happens something and right now those messages are human identified by their name but really need to be identified by their hashes.

as they are, see above.

> Or the dropping of std_msgs/Bool etc.

While I don't necessarily agree with the rationale, this was not really dropped, insomuch as it got 'moved' to `example_interfaces` (https://github.com/ros2/common_interfaces/pull/116).

gramss: @gavanderhoorn thank you for the proper fact checking and updating (at least) me on the current state of some ROS developments. 👍 

We can bring the topic about a ROS2 release-versioned database of common-msgs in the mentioned discourse discussion.

I am open to discuss this topic. The binding should include a guideline how to deal with common messages regardless of the outcome of this discussion.

ros2/README.md
```diff
@@ -0,0 +1,185 @@
...
+## Version
+
+Current version is `0.1.0`.
+
+<a name="server"></a>
+
+## Server Binding Object
+
+This object contains information about the server representation in ROS 2. 
+ROS 2 can use either DDS or Zenoh as its middleware.
```
gavanderhoorn, line 18: Minor perhaps, but while the main RMWs in use may be DDS-based, there are certainly others that aren't (`rmw_zenoh` being a recent addition, but also [eclipse-ecal/rmw_ecal](https://github.com/eclipse-ecal/rmw_ecal) fi).

ros2/README.md
gramss: done

ros2/README.md
gramss: done

ros2/README.md
gramss: I think this is handled as well

ros2/README.md
gramss: Amparo added those as well

ros2/README.md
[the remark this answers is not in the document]
gramss: done

ros2/README.md
gramss: currently this list is not comprehensive. This could be much larger but a better place to look for changes is here: e.g. for Jazzy: [official ROS doc for release changelog](https://docs.ros.org/en/humble/Releases/Jazzy-Jalisco-Complete-Changelog.html)

For an initial PR I would leave it that way and depending of the acceptance and usage of AsyncAPI in ROS this can be extended more and bring back the ROS 2 Version column that way.

Leaving this open to feedback from others for now

ros2/README.md
gramss: We included the design document suggested by @Achllle .

But @fmvilas question is not answered:
>  explain what [this table] is for

I think such an information could be needed to properly integrate/visualize different bindings in one tool. That is why you have your [AsyncAPI message types and formats](https://www.asyncapi.com/docs/reference/specification/v3.0.0#dataTypeFormat:~:text=The%20formats%20defined%20by%20the%20AsyncAPI%20Specification%20are%3A) for.

Right now, this is used as our Rosetta Stone table for the siemens PoC AsyncAPI generator reading ros message file definitions and producing respecting AsyncAPI files.

I hope that the current wording reflects this already. If not, we can make it more distinctive.

ros2/README.md
Achllle: A specific (recent) distribution of ROS 2 + the package name + interface name is the only thing needed to fully define a message. It feels somewhat silly to redefine such messages in the asyncAPI. Without the message headers or import it's not possible (AFAIK) to create the payload like one would do with a json blob. To me the ideal experience would involve linking to the interface definition rather than redefining it. When someone intends to update to a newer distribution, they would have to manually validate all AsyncAPI messages to match the new distribution. Most ROS applications heavily depend on those interface definitions included in ros-base or other common packages, so this would mean a lot of duplication and room for user error.

I don't really see the point with 'black box code'. Such repos will have at least their interface descriptions made available, so whether the code is accessible doesn't really matter.

Perhaps this 'linking' is possible through a binding where if it's a standard message, you'd link to it with:
```yaml
$ref: https://docs.ros.org/en/$ROS_DISTRO/p/geometry_msgs/msg/Twist.html
# OR
$ref: https://github.com/ros2/common_interfaces/blob/$ROS_DISTRO/geometry_msgs/msg/Twist.msg
# OR
$ref: my_custom_interfaces/msg/MyMsg.msg
```
and we include some magic to render the definitions when they're found.

Looking at [NoDL](https://github.com/ubuntu-robotics/nodl), it seems that this approach is happening there as well.

ros2/README.md
gramss: Thank you for your comment!

As you know, there are a quite  acouple of them with my favourite (for the sake of this exact argument) the [christophebedard/rmw_email](https://github.com/christophebedard/rmw_email).

We also had this argument earlier with @fmvilas , that for the moment we only want to focus on DDS and Zenoh based systems. But I think that the majority of the RMW implementations should also be representable through this binding. But this current implication should be put out exactly in this sentence. Therefor, thank you for highlighting it!

Additionally, we are already indicating the vast options one can choose from with ROS2 here: https://github.com/asyncapi/bindings/blob/463fb3d3570cfcb51f5fe609ef348c343e0cbcf2/ros2/README.md?plain=1#L27

ros2/README.md
amparo-siemens: > This seems like a custom message type which is confusing since the [official page](https://docs.ros.org/en/crystal/Tutorials/Topics/Understanding-ROS2-Topics.html#ros2-topic-list) uses [geometry_msgs/Twist](https://docs.ros.org/en/rolling/p/geometry_msgs/msg/Twist.html)

> If you mean the latter (geometry_msgs/Twist vs TwistMsg) then we need to ask @amparo-siemens .
Could you please check if this is just a typo or has something to do how we create the references to messages.
Personally, I believe this is just a typo and agree to @Achllle that this needs to be fixed. Thanks for pointing it out!

Both of you are totally right. In this example TwistMsg looks like a custom msgs and we want to avoid that for sure. Even in the last commit that @gramss did it will be missunderstood because it is just "Twist" and not "geometry_msgs/Twist". This was my mistake because I was trying to put everything in one file (changing the references to other asyncAPI files) in this example so anyone could copy and paste the example and see directly the final appearance of the asyncAPI preview.

When we have several files to define the asynAPI it looks like this.

├── interfaces
│     ├── geometry_msgs.yaml
│     ├── std_msgs.yaml
│     ├── <custom_pkg_msgs>.yaml
│     └── ....
└── head-asyncapi.yaml


head-asyncapi.yaml
```yaml
asyncapi: 3.0.0
info:
  title: Turtlesim example for ROS 2
  version: 1.0.0

servers:
  ros2:
    host: none
    protocol: ros2
    protocolVersion: humble
    bindings:
      ros2:
        rmwImplementation: rmw_fastrtps_cpp
        domainId: 0

channels:
  CmdVel:
    address: /turtle1/cmd_vel
    messages:
      Twist:
        $ref: ./interfaces/geometry_msgs.yaml#/components/messages/Twist

operations:
  CmdVel:
    action: receive
    channel:
      $ref: "#/channels/CmdVel"
    bindings:
      ros2:
        role: subscriber
        node: /turtlesim
```
./interfaces/geometry_msgs.yaml
```yaml
asyncapi: 3.0.0
info:
  title: geometry_msgs
  version: 1.0.0
components:
  messages:
    Twist:
      tags:
        - name: msg
      payload:
        type: object
        properties:
          linear:
            $ref: "#/components/messages/Vector3/payload"
          angular:
            $ref: "#/components/messages/Vector3/payload"
    Vector3:
      payload:
        properties:
          x:
            type: number
            format: double
          y:
            type: number
            format: double
          z:
            type: number
            format: double
```
Our approach is to have it like this, so (at least in my opinion) it is clear that in this case Twist is from geometry_msgs.
For this README we could put it like above so it is clear that "Twist" is from geometry_msgs and knows the structure for the files to be able to see the AyncAPI preview with the VSCode extension for example.

@Achllle, do you think that like this will be clear?

ros2/README.md
amparo-siemens: @Achllle First, thank you for your thoughtful feedback on this discussion. I appreciate your expertise in ROS2 and would like to address some of the points you raised.
 
### **Regarding References in AsyncAPI**
I understand your suggestion about using direct URL references to ROS2 message definitions. While this approach seems intuitive from a ROS2 perspective, the AsyncAPI specification has a specific way of defining message payloads, which relies on the $ref keyword to reference reusable schema definitions. From my understanding, it's not possible to use a direct URL in this context
The $ref syntax you proposed would require custom extensions or resolvers that aren't part of the standard AsyncAPI functionality. This could create compatibility issues with existing AsyncAPI tools and validators.
 
### **On Message Definition Duplication**
You make an excellent point that a ROS2 distribution + package name + interface name is sufficient to fully define a message. What might seem like redundant redefinition actually serves an important purpose in our approach.
By including complete message definitions within the AsyncAPI document, we enable:

- **Automated Code Generation:** Tools can generate client/server code, blocks for low code tools as behavior tree or node red, directly from the AsyncAPI document if it contains all the necessary information.
- **Proper HTML documentation:** Having complete message definitions embedded in the AsyncAPI document ensures that the generated documentation is self-contained and fully informative, allowing even those unfamiliar with ROS2 to understand the API structure and message formats without needing to reference external resources.

 
### **Our Approach: Automation Over Manual Definition**
I believe we might be aligned on the core issue but approaching it differently. Our goal isn't to have developers manually redefine ROS2 messages in AsyncAPI (which would indeed be tedious and error-prone). Instead, we envision automated tools that:

1. Extract message definitions from ROS2 packages
2. Convert them to AsyncAPI format
3. Generate the complete AsyncAPI document with all necessary definitions included

This way, developers don't need to manually maintain these definitions - the tooling handles the synchronization between ROS2 interfaces and AsyncAPI documents

Achllle: @fmvilas can you please comment on 
> would require custom extensions or resolvers that aren't part of the standard AsyncAPI functionality.

specifically whether this would be achievable today through e.g. bindings?

I understand it may be harder to do this, but it's hard to argue that it's also not superior. Automated generation doesn't solve the issue of duplicating information, thus leading to potential conflicts, especially since messages can vary between ROS 2 distros.

> generated documentation is self-contained and fully informative, allowing even those unfamiliar with ROS2 to understand the API structure and message formats without needing to reference external resources.

Documentation generation can take care of linking or even embedding the message definitions so that would address self-containment. That would also solves the issue for those unfamiliar with ROS 2. I'm also not sure why someone unfamiliar with ROS 2 would want to deeply interact with a ROS 2 API without first learning ROS 2 basics.

That being said, if it proves unattainable or would require updating the spec in a major way, I'm fine with the duplication approach, though I might consider using NoDL in that case.

ros2/README.md
Achllle: Yes, that looks significantly better! This slightly more complete example would fit well within the doc itself IMO

ros2/README.md
gramss: agree on getting @fmvilas opinion on this.

I think we cannot alter it from this binding-spec but rather would need to make contributions to the larger asyncAPI body.

@Achllle we should also look at the problem from another angle. ROS already does a great job at defining its messages through a standard approach on the code level with the dedicated `.msg` / `.srv` and `.action` files. Are other supported protocols also doing that way? I was not able to gather such information about MQTT for instance.


---
Regarding your concerns:
>  issue of duplicating information, thus leading to potential conflicts, especially since messages can vary between ROS 2 distros

Can we dig deeper into this?
Are you concerned about the duplication itself or more about the missing reference to the original origin?
The datatypes need to be translated anyway from ROS into AsyncAPI at some point. Either at a rendering stage of HTML or before by holding the information in files.
We could influence this from the ROS UX so that those files are only generated on-demand and deleted afterwards.

If your only concern might be regarding different ros distros, this information is also available at the head asyncAPI file. But this is not pinned on the message level for mix-bistro-robots (that I would like to neglect if possible at a first stage?).

> I'm also not sure why someone unfamiliar with ROS 2 would want to deeply interact with a ROS 2 API without first learning ROS 2 basics.

Reading a documentation of a robot and finding information (where and if e.g. a camera raw stream is made available) is a step before going into the details of a specific application. Could be that a robot offers such information only as a websocket, then one does not need to go deeper into the ROS universe.

Generally, I would love to bring this discussion on a use-case based requirement level. I totally respect your view on this subject but would love to bring it into use-case scenarios that we can either accept or reject (non-relevant) and with it the proposing requirements.
This is something we could focus on the meeting next week.

ros2/README.md
gramss: solved

ros2/README.md
gramss: changes added to the PR

README.md
```diff
@@ -16,6 +16,7 @@ This repository contains the specifications for each AsyncAPI protocol binding.
 * [NATS binding](./nats)
 * [Pulsar](./pulsar)
 * [Redis binding](./redis)
+* [ROS2](./ros2)
```
christophebedard, line 19:
```suggestion
* [ROS 2](./ros2)
```

ros2/README.md
```diff
@@ -0,0 +1,330 @@
...
+      ros2:
+        role: action_client
+        node: /teleop_turtle
+```
+
+<a name="message"></a>
+
+## Message Binding Object
+
+While this object DOES NOT contain any ROS 2 specific properties, it is important to understand how to express the different [ROS2 data types](https://design.ros2.org/articles/legacy_interface_definition.html#:~:text=to%20IDL%20types-,ROS%20type,string,-The%20mapping%20of) in [AsyncAPI message types and formats](https://www.asyncapi.com/docs/reference/specification/v3.0.0#dataTypeFormat:~:text=The%20formats%20defined%20by%20the%20AsyncAPI%20Specification%20are%3A).
```
christophebedard, line 180:
```suggestion
While this object DOES NOT contain any ROS 2 specific properties, it is important to understand how to express the different [ROS 2 data types](https://design.ros2.org/articles/legacy_interface_definition.html#:~:text=to%20IDL%20types-,ROS%20type,string,-The%20mapping%20of) in [AsyncAPI message types and formats](https://www.asyncapi.com/docs/reference/specification/v3.0.0#dataTypeFormat:~:text=The%20formats%20defined%20by%20the%20AsyncAPI%20Specification%20are%3A).
```

ros2/README.md
```diff
@@ -0,0 +1,330 @@
...
+uint16 | integer | uint16
+int32 | integer | int32
+uint32 | integer | uint32
+int64 | integer | int64
+uint64 | integer | uint64
+string | string | string
+array | array | --
+
+It is important to understand that the message header of the AsyncAPI specification does not map with the message header in ROS2.
+For this reason, the AsyncAPI message header will be ignored. If you want to define the header of a message in ROS2, you should put in the payload of the message. [Example](https://docs.ros.org/en/ros2_packages/rolling/api/point_cloud_interfaces/msg/CompressedPointCloud2.html):
```
christophebedard, line 201:
```suggestion
It is important to understand that the message header of the AsyncAPI specification does not map with the message header in ROS 2.
For this reason, the AsyncAPI message header will be ignored. If you want to define the header of a message in ROS 2, you should put in the payload of the message. [Example](https://docs.ros.org/en/ros2_packages/rolling/api/point_cloud_interfaces/msg/CompressedPointCloud2.html):
```

ros2/README.md
```diff
@@ -0,0 +1,330 @@
...
+          $ref: "#/components/messages/uint8/payload"
+        is_dense:
+          $ref: "#/components/messages/bool/payload"
+        format:
+          $ref: "#/components/messages/string/payload"     
+```
+
+## Complete example:
+From the AsyncAPI specification example it could be extracted that:
+- It consist on a ROS2 jazzy application running with Fast DDS as RMW.
```
christophebedard, line 241:
```suggestion
- It consist on a ROS 2 Jazzy application running with Fast DDS as RMW. 
```

ros2/README.md
```diff
@@ -0,0 +1,330 @@
...
+        format:
+          $ref: "#/components/messages/string/payload"     
+```
+
+## Complete example:
+From the AsyncAPI specification example it could be extracted that:
+- It consist on a ROS2 jazzy application running with Fast DDS as RMW. 
+-  `/turtlesim` node is a subscriber of the `/turtle1/cmd_vel` topic and its qos policies.
+- The interface of the `/turtle1/cmd_vel` topic is `Twist` that has a nested type: `Vector3`. Both of them are part of the standard package `geometry_msgs`.
+- `Vector3` has already the types converted to AsyncAPI types and formats (number and double), instead of using the ROS2 types.
```
christophebedard, line 244:
```suggestion
- `Vector3` has already the types converted to AsyncAPI types and formats (number and double), instead of using the ROS 2 types.
```

christophebedard (commented):
ROS2 -> ROS 2

README.md
gramss: thanks! we fixed that right away and look for further typos of that direction

Achllle:
@gramss @amparo-siemens @fmvilas I'm hoping this doesn't get stuck in limbo - what needs to be done to get this over the finish line?

ros2/README.md
fmvilas: I think this conversation can be marked as resolved already, right? Unless you want to provide more information.

ros2/README.md
fmvilas: Feel free to mark the conversation as resolved if it's done.

ros2/README.md
fmvilas: `$ref` can't be used non-JSON structures cause it uses the JSON Reference and JSON Pointer syntax. So I completely discourage the usage of `$ref` for `.html`, `.msg`, etc. We definitely had this discussion before but no one took the lead on that yet: https://github.com/asyncapi/spec/issues/930. The folks using Protobuf also want it so they can reuse `.proto` files.

So I'd leave it out of this PR, work on the proposal I linked above, and come back to this binding to adapt it.

ros2/README.md
fmvilas: Feel free to mark the conversation as resolved if it's done.

ros2/README.md
gramss: I am 100% fine with this.

@Achllle let's keep it like it is, but feel free to join is in asyncapi/spec#930 to take this discussion further.
We also, identified valid usage of having a reference to our `.msg` (etc.) files.
But this is too big for this PR.

@amparo-siemens feel free to push this forward together with the community :)

I'll mark this as resolved for now.

fmvilas:
LGTM as it is 👍 

We should now proceed with the PRs in the github.com/asyncapi/spec and github.com/asyncapi/spec-json-schemas repos. I don't think we have to add anything to the github.com/asyncapi/parser-js repo.

In the spec repo, we should add the `ros2` binding in the Server Object, Channel Object, Operation Object, Message Object, and all the binding references in the Components Object. This is the file to modify: https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#componentsObject. The PR title should start with `feat` because we want this to trigger a new version the spec: 3.1.0.

In the spec-json-schemas repo, we should add the `ros2` binding JSON Schema definition in the same places as above. I'm less familiar with this repo but maybe @jonaslagoni or @Pakisan can guide you a little bit.

And that's it, with that in place we can finally proceed to merge everything have have the `ros2` binding available.

cc @derberg @dalelane @char0n @GreenRover please have a look and let them know if I missed something 🙏

Pakisan:
> LGTM as it is 👍 
> 
> 
> 
> We should now proceed with the PRs in the github.com/asyncapi/spec and github.com/asyncapi/spec-json-schemas repos. I don't think we have to add anything to the github.com/asyncapi/parser-js repo.
> 
> 
> 
> In the spec repo, we should add the `ros2` binding in the Server Object, Channel Object, Operation Object, Message Object, and all the binding references in the Components Object. This is the file to modify: https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#componentsObject. The PR title should start with `feat` because we want this to trigger a new version the spec: 3.1.0.
> 
> 
> 
> In the spec-json-schemas repo, we should add the `ros2` binding JSON Schema definition in the same places as above. I'm less familiar with this repo but maybe @jonaslagoni or @Pakisan can guide you a little bit.
> 
> 
> 
> And that's it, with that in place we can finally proceed to merge everything have have the `ros2` binding available.
> 
> 
> 
> cc @derberg @dalelane @char0n @GreenRover please have a look and let them know if I missed something 🙏 

@gramss 
Yep, just bring schemas back to schemas repo and everything will be fine

Write to me if you will stuck

amparo-siemens:
Is there anything pending here @fmvilas? 😄

derberg:
well schemas should not be released as bindings and spec is not released 😃 for that we need new release cycle to start, and volunteer release coordinator: https://github.com/asyncapi/spec/blob/master/RELEASE_PROCESS.md

gramss:
@fmvilas there might be another repo waiting for this to become official.. 😇 
[ROSITA (ROS Siemens Introspection Tool for AsyncAPI)](https://github.com/siemens/rosita)

Pakisan:
> @fmvilas any update on the V3.1 minor release? 🙂 

Release on me, my bad. I'll prepare needed things in a few days

derberg:
in case you missed https://github.com/asyncapi/spec-json-schemas/pull/582#issuecomment-3347636902

@amparo-siemens @gramss I'm available for onboarding call. Please reach out to me at lpgornicki@gmail.com to sync on the exact day and time. It's purely bindings maintainance onboarding but I'll also share overall info about how the AsyncAPI Initiative operates.

derberg:
Hey folks, fyi we scheduled our Spec Triage meeting: https://github.com/asyncapi/spec/issues/1131#issuecomment-3533564832 

First time in a very, very long time 😄 I think v3-trauma has passed 😄

CODEOWNERS
```diff
@@ -17,3 +17,4 @@
 /pulsar/       @VisualBean
 /sns/          @dpwdec @iancooper
 /sqs/          @dpwdec @iancooper
+/ros2/         @amparo-siemens @gramss
```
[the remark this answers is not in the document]
amparosancho: Okay, @gramss could you put this one? 🙂

fmvilas (approved):
LGTM 👍

